PR #15, opened 2019-04-25
Add CircleCI config

fgimenez:
Sets up CircleCI config with a build test and adds build badges to the main page.


burdges:
I'm not familiar with this thing, but looks reasonable.  :)

fgimenez:
Thanks! :)

PR #16, opened 2019-04-26
Add link checker

fgimenez:
Adds a link checker step to the CI build, the site is served locally and all the links are checked, if any of them is broken (even src fields for scripts) the build fails.

I've taken the opportunity to fix a broken link to NPoS and a broken script src (the latter comes from mkdocs-material).


fgimenez:
Just pushed some additional changes that recover the missing javascript file (present on mkdocs-material repo), and with this the search box works again.

This PR was proposed so that people in the research team is aware of this changes, merging now.

PR #45, opened 2019-07-25
Proofread Slashing

lsaether:
The write-up uses capitalized Grandpa and Babe instead of acronyms GRANDPA and BABE. Should these be adapted?


lsaether:
Should [this](https://github.com/w3f/research/pull/45/files#diff-6fc35bdd6f9fb6a0a907021a7c837afeL108) be changed from **level 3** to **level 2**?

> We consider it a misconduct of level 1 if the slashing fraction is at most 1%, and of **level 2** otherwise.

burdges:
I went ahead and merged but https://github.com/w3f/research/pull/45/commits/5ab48ab94209154c8adb905b353b8563f5a208b5 should be reviewed @AlfonsoCev although looks right to me

PR #49, opened 2019-09-05
Add research team members

fgimenez:
This will add empty documents for each of the team members and links on the left side menu. 

@FatemeShirazi is it ok to add the document of each researcher empty? If so we can merge this and add content later, let me know.


fgimenez:
Thanks!

PR #54, opened 2019-10-10
delete because of redundancy

hndnklnc:


burdges:
I'll deal with these.

PR #58, opened 2019-10-29
Partial restructure of networking overview

infinity0:
Mostly for my understanding coming in from being an outsider, but also hopefully make the document a bit more self-contained and future proof.


docs/polkadot/networking/1-overview.md
```diff
@@ -37,16 +36,16 @@ abstract part of the process of executing transactions:
    the internals of which are outside the scope of Polkadot, chosen by each
    parachain for themselves.
 
-3. Parachain block attestation: collators also generate additional data and
-   pass this to the parachain validators. The ultimate aim of this data is for
+3. Parachain block attestation: collators also generate additional data (called PoV) and
+   pass this to the parachain validators. The ultimate aim of the PoV is for
```
infinity0: The reason I called it "the data" before was trying to make this generic and independent of any detailed potential future changes in grandpa, e.g. if it is decided to have multiple types of PoVs or something else.

PR #65, opened 2019-11-19
store CI artifacts

fgimenez:
Closes https://github.com/w3f/infrastructure/issues/45


fgimenez:
This is how it looks like https://489-162593780-gh.circle-artifacts.com/0/root/project/site.tar.gz

infinity0:
Ah, I actually meant unzipped so it can be browsed on the web. But it's OK, I realised that we have [gh-pages here](https://w3f.github.io/research/) so I just built it manually and pushed there.

PR #66, opened 2019-11-21
Expand and update networking section of overview paper

infinity0:


infinity0:
The summary table at the end of the section could also be updated but I wanted to think for a bit on the best way to do that.

PR #73, opened 2020-02-13
First draft of the glossary (WIP)

lamafab:
**UPDATED 2020-02-18 10:50 CET**
Removed:
- CD
- CG
- CP
- CQ
- Kademlia
- local clock
- Phragmén
- Collators (plural, add to `Collator` description)
- Nominators (plural, add to `Collator` description)
- Validators (plural, add to `Collator` description)
- Candidate validators

Adjusted
- BABE description
- Nominator description
- Validator description

I wouldn't make those descriptions too long. I think this should be fine. Also, we can link to https://wiki.polkadot.network/docs/en/glossary .

Printscreens:
image
image

**ORIGINAL**
Hi

Here's the first draft of the glossary. I've also added some labels to certain parts in the document and reference those. I'd appreciate some feedback before I'll try to resolve the merge conflicts...

**NOTE**: I've left out the following entries, since those are quite common in the computer/crypto space and not directly specific to Polkadot. Let me know if some of those should be included (or at least those should have external references inside the document?).
- Blob
- BLS signatures
- Byzantine
- Casper FGG
- Casper TFG
- Consensus
- DHT
- ed25519
- GHOST
- HDKD
- MAC
- Merkle proof
- Merkle tree
- PRF
- RPC
- schnorrkel/sr25519
- State
- State transition
- Wasm
- XOR
- ZKP

**Printscreens of the glossary**
image
image
image


burdges:
I always call blob a candidate blob in AnV to be more clear.  It's exactly the same as a PoV block.

lamafab:
> I always call blob a candidate blob in AnV to be more clear. It's exactly the same as a PoV block.

@burdges By "blob" I was thinking the Runtime ("Wasm blob"), which is defined as *binary file* AFAIK.

AlfonsoCev:
I think the glossary should contain terms that are either 
a) specific to Polkadot, or
b) used across several section and not re-defined every time.

Since neither is the case for the following terms, I think we should remove them: blockchain, block, CD, CG, CP, CQ, Kademlia, local clock, Phragmén.

On the other hand, I think we should expand some descriptions in the glossary, such as for BABE, GRANDPA, XCMP, validator, collator, nominator, etc.

Finally, I agree with Fabio about leaving out the extra terms (blob, BLS signatures, etc.), but we should make sure to add a description and/or reference for each of them in the section they are introduced. That way we don't lose the reader who doesn't know these terms.

FatemeShirazi:
Thanks a lot, Fabio, it's a great start. I think this glossary should be about giving an overview of terms, roles, and concepts that we came up with. Hence, leaving out the terms that you suggested makes sense of course.  

I have some small comments though. I agree with Alfonso, Phragmen and Kademlia don't need to be there and some terms need more explanation like nominators. I would also merge the plural version of the nominator and collator within the same line just have plurals in () for example.

lamafab:
@AlfonsoCev @FatemeShirazi Okay, I will adjust that.

docs/polkadot/Polkadot-OverviewPaper/background.tex
```diff
@@ -8,45 +8,32 @@ \section{Glossary and Background}
...
+    PoV & \emph{Proof-of-Validity} - Mechanism where a validator can verify a block without having its full state. && \ref{sec:parachainblockproduction} \\
+    PoW & \emph{Proof-of-Work} - Mechanism where parties vote with processing power. && \\
+    Relay\newline- Chain & Ensures global consensus among parachains. && \ref{sec:relaychain} \\
+    Runtime & The Wasm blob which contains the state transition functions, including other core operations required by Polkadot. && \ref{par:state_transition} \\
+    Sentry\newline- nodes & Specialized proxy server which forward traffic to/from the validator. && \\
+    STVF & \emph{State-Transition-Validation-Function} - A function of the Runtime to verify the PoV. && \ref{sec:parachainblockproduction} \\
+    Validator & The elected and highest in charge party who has a chance of being selected by BABE to produce a block. A set of candidate validators is defined as \Can . The number of validators to elect is defined as \nval . & \val & \ref{par:validators} \\
+    VRF & \emph{Verifiable-Random-Function} - Cryptographic function for determining elected validators for block production. && \ref{sec:babe} \\
```
burdges: We use VRFs for many things.  You'll find a better but too long definition in https://github.com/w3f/research/blob/master/docs/papers/sass/sass-1-vrfs.tex

PR #77, opened 2020-03-18
A&V networking initial proposal

infinity0:
Imported from HackMD.

This PR is just for convenience for others to make comments.

It deletes some previous content which will be re-integrated before merge.


docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+### Special considerations
+
+S1: A small fraction of nodes may have poor reachability, and need to communicate via a proxy. Otherwise, we can assume the graph of communication links is mostly-connected via the public internet.
+
+This is an optional consideration; we believe we do not need to consider it in the initial implementation, and so our solution for this is described in a separate section in this document. Parts of the main proposal that will need to be changed or extended for this, are marked [TBX S1 #($ref)], TBX standing for "to be extended".
+
+## Protocol overview
+
+We rely on an underlying gossip network that allows us to broadcast various metadata to everyone, namely:
```
mxinden:
```suggestion
We rely on an underlying gossip network that allows us to broadcast various metadata to every node of the relay chain, namely:
```

Parachain nodes don't need to receive this, right?

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+1. receipts of specific pieces ("I have X piece")
```
mxinden:
```suggestion
1. receipts of specific pieces ("I have pieces X, Y and Z")
```

docs/polkadot/networking/3-attestation.md
mxinden: Do I understand correctly that for each period each validator gossips one metadata message? Thus with 1_000 validators there would be 1_000 meta data gossip messages per period?

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+2. receipts of "enough" pieces ("I have >1/3 of pieces, I don't need more"). This could also double as an attestation to everyone else, that there is availability from the attester's position on the network.
+3. [TBX S1 #1]
+
+This metadata should be gossiped every few seconds. The data of the actual pieces are distributed via a separate topology as described below:
+
+Recall that we have a disjoint partition of N validators into C sets of parachain validators, each set having size N/C. For our purposes for this subprotocol, we will randomly assign a co-ordinate (c, i) to every validator, with c in [0, C) and i in [0, N/C). Fixing c and varying i defines a particular parachain validator set; varying c and fixing i defines what we'll call a particular validator "ring". This name is only meant to be very slightly suggestive, the precise structure and its justification will be described below.
+
```
mxinden:
```suggestion

Example:

Let's say we have 20 validators `[a, b, c, ..., t]` and 5 parachains. The co-ordinates of the validators could look like such `[a: (0, 0), b: (0, 1), c: (0, 2), d: (0,3), e: (1, 0), ..., t: (3, 3)]`. The validator set of the first parachain would be `[a, b, c, d]`. The first validator ring would be `[a, e, i, m, q]`, the second `[b, f, j, n, r]`.

```

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+We rely on an underlying gossip network that allows us to broadcast various metadata to everyone, namely:
+
+1. receipts of specific pieces ("I have X piece")
+2. receipts of "enough" pieces ("I have >1/3 of pieces, I don't need more"). This could also double as an attestation to everyone else, that there is availability from the attester's position on the network.
+3. [TBX S1 #1]
+
+This metadata should be gossiped every few seconds. The data of the actual pieces are distributed via a separate topology as described below:
+
+Recall that we have a disjoint partition of N validators into C sets of parachain validators, each set having size N/C. For our purposes for this subprotocol, we will randomly assign a co-ordinate (c, i) to every validator, with c in [0, C) and i in [0, N/C). Fixing c and varying i defines a particular parachain validator set; varying c and fixing i defines what we'll call a particular validator "ring". This name is only meant to be very slightly suggestive, the precise structure and its justification will be described below.
```
mxinden: > we will randomly assign a co-ordinate

How does randomly assigning co-ordinates to validators look like? Who does this? How is this view consistent across validators?

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+## High-level requirements
+
+Ensure that ceil(N/3) of the pieces of all C blocks remains available and retrievable for a reasonable amount of time, across the N validators.
+
+## Design considerations
+
+### General aims
+
+A1: Reduced redundancy & latency: pieces should (mostly) only be sent to the nodes that need it, i.e. its main storer, and the secondary checkers.
```
mxinden: Can you elaborate what a *checker* is here? As far as I can tell https://research.web3.foundation/en/latest/polkadot/Availability_and_Validity.html does not use this term.

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+3. [TBX S1 #1]
+
+This metadata should be gossiped every few seconds. The data of the actual pieces are distributed via a separate topology as described below:
+
+Recall that we have a disjoint partition of N validators into C sets of parachain validators, each set having size N/C. For our purposes for this subprotocol, we will randomly assign a co-ordinate (c, i) to every validator, with c in [0, C) and i in [0, N/C). Fixing c and varying i defines a particular parachain validator set; varying c and fixing i defines what we'll call a particular validator "ring". This name is only meant to be very slightly suggestive, the precise structure and its justification will be described below.
+
+A validator ring is mostly-connected as permitted by the physical topology. Nodes within this ring talk to each other periodically via short-term and low-cost QUIC connections.
+
+For a given parachain, the primary checkers are also mostly-connected, as permitted by the physical topology. Likewise the secondary checkers are also mostly-connected.
```
mxinden: What is a primary checker? This term is never used before and not part of https://research.web3.foundation/en/latest/polkadot/Availability_and_Validity.html.

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+P1SA  | Y            | N
+P1SB  | Y            | Y
+P2SA  | Y            | N
+P2SB  | N            | Y
+
+Note that there is also background activity, as described below.
+
+## Protocol phase 1: initial distribution
+
+Every validator is both a distributor of C pieces and a distributee (recipient) of C pieces [TODO: correction, explanation]. Every piece has one source parachain and one main target storer, and so we can index pieces with a tuple (c_s, (c_t, i_t)).
```
mxinden:
```suggestion
Every validator is both a distributor of C pieces and a distributee (recipient) of C pieces [TODO: correction, explanation]. Every piece has one source parachain and one main target storer, and so we can index pieces with a tuple (c_s, (c_t, i_t)) which would read as *the piece with source parachain c_s and destination validator on parachain c_t with index i_t*.
```

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+In phase 1, pieces are distributed by the source parachain validators (c_s, \*) to the main target storers. This happens in two stages. Stage A is where most of the material is distributed, and stage B acts as a backup mechanism for anything that was missed during stage A.
+
+**Stage A**
+
+As a distributor, each validator (c, i) attempts to send the relevant pieces for their parachain's block (c, (c', i)), to everyone else on their ring (c', i) for all c' != c. Conversely as a distributee, each validator expects to receive their relevant pieces (c', (c, i)) for other parachains' blocks, from everyone else on their ring (c', i) for all c' != c.
```
mxinden: > to everyone else on their ring (c', i) for all c' != c.

If I understood the concept of a ring correctly, there is ever only one validator per parachain on such ring. Thus `c' != c` is not needed here, right?

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+In more detail:
+
+Each distributor (c, i) will, with parallelism = C / 4, for s in [0..C), try to send the relevant piece to target t = (c+s, i) [TBX S1 #2]. C / 4 comes from our estimate that T_b ~= 4 * T_L.
```
mxinden: Small nit pick:

```suggestion
Each distributor (c, i) will, with parallelism = C / 4, for s in [0..C), try to send the relevant piece to target t = ((c+s) %C, i) [TBX S1 #2]. C / 4 comes from our estimate that T_b ~= 4 * T_L.
```

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+Trials are done with a timeout, slightly larger than T_l. Sending is via QUIC. In order for it to be treated as a success, it should include an acknowledgement of receipt. Note this is orthogonal from the gossiped receipts which include a validator signature; by contrast this transport-level receipt can be assumed to be already protected by QUIC transport authentication.
```
mxinden: > by QUIC transport authentication

As discussed in the past today there is no notion matching chain identity with network identity. I remember you had some design document for authentication. Maybe worth linking here.

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+If a gossiped receipt is received at any point during the whole process, for a target for a piece, then we can interpret that to mean that the target obtained the piece from a different sender in the meantime, and we should cancel the sending attempt with success.
+
+**Stage B**
+
+As a distributee, if after a grace period we still haven't received our piece (c', (c, i)) from a parachain validator (c', i), then we will ask the other validators (c', i') for all i' != i in that parachain for the piece (c', (c, i)), load-balanced as described in more detail below.
+
+This gives the distributee a more direct level of control over obtaining their own pieces.
+
+As a distributor, if after our own stage A process is finished, we have received fewer than ceil(N/3) of the receipts of another ring i' - i.e. receipts for piece (c, (c', i')) from peer (c', i') for all c' - then we will begin the stage A process for the other ring i' too, load-balanced as described in more detail below.
```
mxinden: I am confused. The *receipt* here is a *Stage A* receipt or a gossip receipt? In the former case why would a validator expect to receive receipts of validators from a different ring?

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
+
+Each distributor (c, i) will, with parallelism = C / 4, for s in [0..C), try to send the relevant piece to target t = (c+s, i) [TBX S1 #2]. C / 4 comes from our estimate that T_b ~= 4 * T_L.
+
+Trials are done with a timeout, slightly larger than T_l. Sending is via QUIC. In order for it to be treated as a success, it should include an acknowledgement of receipt. Note this is orthogonal from the gossiped receipts which include a validator signature; by contrast this transport-level receipt can be assumed to be already protected by QUIC transport authentication.
+
+If a gossiped receipt is received at any point during the whole process, for a target for a piece, then we can interpret that to mean that the target obtained the piece from a different sender in the meantime, and we should cancel the sending attempt with success.
+
+**Stage B**
+
+As a distributee, if after a grace period we still haven't received our piece (c', (c, i)) from a parachain validator (c', i), then we will ask the other validators (c', i') for all i' != i in that parachain for the piece (c', (c, i)), load-balanced as described in more detail below.
```
mxinden:
```suggestion
As a distributee, if after a grace period we still haven't received our piece (c', (c, i)) from a parachain validator (c', i), then we will ask the other validators (c', i') for all i' != i (from all other rings) in that parachain for the piece (c', (c, i)), load-balanced as described in more detail below.
```

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
-**Authors**: Rob Habermeier
+**Authors**: Ximin Luo
 
-**Last updated**: 12.09.2019
+**Last updated**: 2020-03-18
 
 ====================================================================
-## Parachain networking
 
-### Attestation Gossip: gossiping parachain blocks
```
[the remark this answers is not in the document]
infinity0: I addressed this in the PR summary:

> This PR is just for convenience for others to make comments.
>
> It deletes some previous content which will be re-integrated before merge.

docs/polkadot/networking/3-attestation.md
```diff
@@ -1,54 +1,268 @@
...
-        - Relay Chain Parent Hash
-    - This is an attestation on the candidate with given hash, that indicates the candidate is both valid and has block data available. May only be sent to peers who are aware of the `Candidate`.
-- **Invalid:**
-    - Members
-        - Candidate Hash
-        - Validator public key
-        - Signature
-        - Relay Chain Parent Hash
-    - This is an attestation against the candidate with given hash. May only be sent to peers who are aware of the `Candidate`.
+Let T_B represent the average time it takes for a distributor to transfer all pieces of a block, i.e. a single piece takes roughly T_B/N. This is only the bandwidth and not latency, so only if T_B/N is negligible then the successful direct transfer of a single piece including acknowledge takes about 1 RTT.
```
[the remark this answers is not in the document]
infinity0: > _N_ is used both for the number of validators as well as the number of erasure encoding pieces. The above suggestion should make it more obvious what is meant here.

N is defined in both these ways in the "Background" section already, do you think it's not clear from there?

> I am having difficulties understanding this sentence. The time something takes (e.g. T_B) is equal to its latency, not to its bandwidth.

The time something takes to transfer, is the link latency (time) plus its size multiplied by the bandwidth (in size per time). Only if the latter is negligible, does the transfer time actually equal the link latency.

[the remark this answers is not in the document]
infinity0: But this isn't the case - with our current numbers, a single piece (30KB) might take up to 10-20 times the MTU (1500 bytes).

Nevertheless, if network conditions are good you can send out 10-20 MTU-sized packets at once, without waiting for an ack, and then the peer can ack all of these packets at once. The bandwidth is small so latency dominates, and this effectively will take ~1 RTT.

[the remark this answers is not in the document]
infinity0: I'll try to add this sort of explanation into a footnote to clarify what is meant here, though.

docs/polkadot/networking/3-attestation.md
infinity0: Thanks, I've tweaked % to mod, to be consistent with the rest of the doc

docs/polkadot/networking/3-attestation.md
infinity0: Thanks for the example, I made a minor fix - the co-ord for `t` should be `(4, 3)`.

docs/polkadot/networking/3-attestation.md
infinity0: It's implied, but I thought it was clearer to be explicit about it.

docs/polkadot/networking/3-attestation.md
infinity0: It's a gossip receipt. I'm not sure what you mean by a "stage A" receipt, these gossip receipts are the same whether sent in stage A or B.

There is a transport-level ACK to confirm a direct-send was received correctly but that's a separate idea and only mentioned in the lower-level context of a QUIC connection.

infinity0:
Thanks for the comments! I've merged some of the suggestions, the other stuff I will take a little longer to go through.

docs/polkadot/networking/3-attestation.md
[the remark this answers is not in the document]
mxinden: > N is defined in both these ways in the "Background" section already, do you think it's not clear from there?

I would suggest applying my suggestion. Not mandatory, just a suggestion.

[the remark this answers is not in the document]
mxinden: > The time something takes to transfer, is the link latency (time) plus its size divided by the bandwidth (size per time). Only if the latter is negligible, does the transfer time actually equal the link latency.

This makes more sense now. Thanks for explaining.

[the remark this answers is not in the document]
mxinden: > But this isn't the case - with our current numbers, a single piece (30KB) might take up to 10-20 times the MTU (1500 bytes).

These assumptions are helpful. Would be good to have them included in the document as well.

docs/polkadot/networking/3-attestation.md
mxinden: :+1: good to know that I am not missing something. Let's leave it as it is.

docs/polkadot/networking/3-attestation.md
mxinden: Understood. How about the following chance to make it more clear?

```suggestion
As a distributor, if after our own stage A process is finished, we have received fewer than ceil(N/3) of the receipts via the gossip overlay of another ring i' - i.e. receipts for piece (c, (c', i')) from peer (c', i') for all c' - then we will begin the stage A process for the other ring i' too, load-balanced as described in more detail below.
```

infinity0: I don't think this is necessary - "receipt" and "gossiped receipt" are used interchangeably many times already in this document, and I don't see much point fixing all 20+ usages of it one way or the other.

docs/polkadot/networking/3-attestation.md
infinity0: Fixed in the latest commit.

docs/polkadot/networking/3-attestation.md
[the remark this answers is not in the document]
infinity0: Fixed in the latest commit.

docs/polkadot/networking/3-attestation.md
infinity0: I added a paragraph about this to the latest version, please take a look.

docs/polkadot/networking/3-attestation.md
infinity0: I've added a part describing this to "high-level requirements" (near the top).

We've started using the terms "preliminary checker" and "approval checker" instead of primary vs secondary checker, and I've update this document accordingly.

docs/polkadot/networking/3-attestation.md
infinity0: (explained in my other comment)

infinity0:
@mxinden I've dealt with your other comments I hope, please take a look.

infinity0:
I haven't yet been briefed on sentry nodes and I was told that it is a rather ad-hoc idea, and there is no single design behind it, and different people have different ideas on what security purpose they are supposed to achieve. So we certainly need to talk about that in a more principle way. How about a video call next week?

> we want to stick with the current sentry node setup, we would need to stick with gossiping as our only communication channel.

I really don't think gossip will be able to handle the amount of traffic that the higher-level A&V protocol is expected to need in the mid-term.

mxinden:
@infinity0 I am referring to this document quite often. Referring to a file in this repository rather than a pull request would be easier. What do you think of either merging this with a *work-in-progress* flag above it, or merging parts of it? Would also ensure we don't have many long lived pull requests.

infinity0:
@mxinden ok, I will merge the earlier attestation gossip stuff from Rob back in (I understand it much better now), merge this PR and then deal with the rest of the stuff (basically sentry node proxies) later in a separate PR. Should be able to do this on Friday, since I have to prepare for the KE talk on Thursday, sound good?

mxinden:
Sounds good. Thanks!

infinity0:
Merged!

- I've kept Rob's attestation gossip as-is in a separate document - after thinking about it it's more closely related to the block-production process, and less related to A&V. However it could generally do with a touch-up and I'll go over it again later.
- I've added the remaining items as a TODO in the document itself. Will pick it up immediately next work.

PR #78, opened 2020-03-24
Fix dead link on BADASS BABE page

FlorianFranzen:
Simple fix for a dead link on research.web3.foundation


swasilyev:
It used to be a link to our academic article (no located at https://github.com/w3f/research/blob/master/docs/papers/sass/sass.tex), but you changed it to some earlier write-up. We will fix it as soon as we have a draft.

PR #81, opened 2020-04-07
Xcmp suggestions leon

LeonGB:
Several rewrites to the XCMP document


AlistairStewart:
So mostly good, but I noted the following:

Problemm definition: the relay chain does not store messages, it only authenticates them. 

Messages need to be authenticated via the relay chain so we can be sure that the messages we act on are the same as those sent. In more detail, a relay chain fork defines a history of all the parachains and we want to make sure that what messages a parachain recieves are consistent with those other parachains sent to it in that history. The question then is how we authenticate them efficiently.

Goals: For (1), all validators may need to be able to check the validatity of a PoV block for some time, not just the parachain validators. Full nodes of the parachain, including fishermen, need to be able to check the parachain block's validity at the time without having the PoV block. And it's not even clear that we can do the latter easily now.

Para, and honestly this is a name I've seen Gav use, refers to parachain or parathread, for times when parachain would be confusing, since properly a prarthread is not a full parachain. This is not necessarily the best terminology, but we do need to differentiate parathreads, full parachains and things that might be either.

PR #84, opened 2020-04-30
Introduce an Implementors' Guide to describe the software architecture and design...

rphmeier:
...of a "Parachain Host". In our case, this is the Relay chain.

This will describe the processes, functionality, and challenges of an implementation of a Parachain Host in the context of the other documentation in this repository.

What this is not:
 - A specification of every aspect of a Polkadot node or the runtime. 
 - A research document describing a protocol.

Those two categories of documents describe functionality, but don't cover the way that the functionality is implemented. This document is aimed to fill that gap.


rphmeier:
@burdges Maybe the AnV stuff would go better as its own PR?

rphmeier:
Architecting the "core" behavior seems pretty straightforward, but it's not that clear how each of the other components should look.

**Collator Logic**
 Collator logic isn't yet covered by the guide. It should consist entirely of core and networking processes. I'd lean towards refactoring them to be included in the same crate as the validator logic, but where the overseer can be configured to run as a collator instead of / in addition to as a validator.

**Block-based work**

Many processes are expected to be run together within the context of a particular block and should be loosely coupled. For example, this non-exhaustive set of processes:
  - Issuing backing statements for candidates for inclusion in a child of relay parent block B
  - Circulating those backing statements via gossip (should be pinned to the same block B)
  - Collating a candidate for inclusion in block with relay parent B
  - Circulating the collation to validators at relay parent B
  - Seconding a candidate at ...
  - Authoring a new relay chain block on top of parent B
  - Circulating availability-pieces that are relevant at parent B
  - Issuing availability bitfields that are relevant at parent B
  - Circulating those bitfields...

All of these tasks should be spun up at the same time (i.e. after import of B) because they represent all the work that the validator should be doing at that time. They should also all presumably be torn down at the same time. They also often need bidirectional communication with each other - core processes usually have a networking counterpart that they send and receive information through. Some core processes need to communicate with each other. The block authorship process needs to get backed candidates from the candidate backing process, the availability bitfield circulation process, etc.. The candidate backing process needs to learn which candidate is seconded. So we need to support fairly flexible messaging between processes.

In the past, I used the term "leaf-work" as a catch-all to describe this, but I think "block-based" work is a better term as the block we are building on is not necessarily a leaf. I am thinking about a specific situation where an invalid parachain candidate forces a rollback, so the block we are building on may have children which we are no longer allowed to follow.

**Networking**
The architecture of the networking tasks right now will probably heavily hinge on shared components between the networking processes. We should be able to swap out the entire networking implementation as well as individual networking processes with mocked or alternative versions. 

This implies that we'd want a "networking overseer" process as a counterpart to the "core overseer". Right now, we'd implement most networking processes (statement circulation, PoV-block distribution, availability pieces distribution) using `substrate-network-gossip`. 

It's unclear now whether it's better to create a separate instance of the gossip handler for each networking process or to have a single shared instance of the gossip handler across all the networking subprocesses. Having multiple gossip handlers means duplicating our "neighbor packets" - those packets that we send to peers to tell them about our protocol state, and which information we'll accept. Having a single gossip handler means that the message-handling logic for different pieces of information will be inextricably tied together.

**Processes and Jobs**

Previously I've been thinking of processes as relatively short-lived work that are all spawned by an overseer. But the considerations of this comment are making me think that a different approach, where the overseer spawns long-lived **Processes** that internally spawn short-lived **Jobs**, suits our requirements better.

1. Jobs speak to each other via their processes, and processes speak to each other via their overseer. This hierarchy of communication will help eliminate race conditions.
2. Processes will spawn jobs based on messages from their overseer. The overseer will determine that it is time to author on relay parent block B, and will then send an `OverseerMessage::AuthoringOn(B)` to all processes. Those processes will then spawn a job to perform work localized to block B. Likewise, the overseer will send an `OverseerMessage::AuthoringConcludedOn(B)` to all processes when it is time to close the authoring process on B (i.e. when a child of B has been imported). We may merge these messages. When Process X starts authoring on B and wants to send a message to Process Y, a `ProcessYMessage::SomeData(_)`, that message must pass through the overseer and thus cannot arrive at Process Y before the `AuthoringOn` message, eliminating a race condition. Likewise, the overseer process doesn't require logic describing how to launch each job. It just needs to launch the processes on startup, and afterwards serves only as an authoring signal and message bus.
3. APIs of processes should be separate from implementation. The set of messages that a process will accept should not depend on the implementation of that process. If this rule is followed, the protocols can be defined in separate crates from the implementations, which will make the codebase more modular.

rphmeier:
(reposting from chat onto GitHub).

What I've been calling the node-side behavior is a lot more clear to me than how the runtime should be architectured. First, I don't want to deal with the behavior of the parachain slot auctions, crowdfunding, etc. but only deal with an API that allows parachains/parathreads to be registered and deregistered, scheduled or unscheduled.
The functionality that I'd be looking to rearchitect comes from the registrar and parachains modules. And the attestations module which is kind of a kludge. I find these modules to be monolithic, potentially bug-prone, and hard to follow. Ideally we'd have ways to break these down to smaller pieces or interfaces that aren't so tightly coupled to each other
The runtime is responsible for most of the hard stuff - validator duty assignments, code upgrades, tracking inclusion state of parachain blocks, parathread scheduling. I'm not sure how to break it down but I'll spend some time thinking about it today. Races between governance-induced Calls and our modules' on_initialize logic is something I'm really worried about.

It would help to first write out the series of requirements of the runtime, so we can architect out how to carry out those tasks.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,342 @@
...
+* Network - a mockable interface for receiving and submitting signed statements about candidates.
+
+If the local key is not present or there is no local assignment, the candidate process will run in an observer mode where it will witness candidates that other validators backs without participating in the process itself.
+
+**Operation**
+
+Pseudocode:
+```rust
+loop {
+  let signed: SignedStatement = network_statements.next()?;
```
mxinden, line 192:
```suggestion
  let signed: SignedStatement = network_statements.next().await?;
```

While hopefully all logic within this realm can be synchronous, communication to the outside world (or to the lower Substrate levels) will need to be asynchronous.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,342 @@
...
+struct SignedStatement {
+  statement: Statement,
+  signed: ValidatorId,
+  signature: Signature
+}
+```
+
+----
+
+## Glossary
```
mxinden, line 307: Why not link and extend https://substrate.dev/docs/en/overview/glossary ? (Why yet another Glossary ? :yum: )

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,342 @@
...
+1. The collator forwards the candidate and PoV to validators assigned to the same parachain via the Collation Distribution Process.
+1. The validators assigned to a parachain at a given point in time participate in the Candidate Backing Process to validate candidates that were put forward for validation. Candidates which gather enough signed validity statements from validators are considered "backed" and are called backed candidates. Their backing is the set of signed validity statements.
+1. A relay-chain block author, selected by BABE, can include up to one (1) backed candidate for each parachain to include in the relay-chain block alongside its backing.
+1. Once included, the parachain candidate is considered to be "pending availability". It is not considered fully included until it is proven available.
+1. In the following relay-chain blocks, validators will participate in the Availability Distribution Process to ensure availability of the candidate. Information regarding the availability of the candidate will be included in the subsequent relay-chain blocks.
```
mxinden, line 62:
```suggestion
1. In the following relay-chain blocks, validators will participate in the Availability Distribution Process (see https://github.com/w3f/research/pull/77 for network related details) to ensure availability of the candidate. Information regarding the availability of the candidate will be included in the subsequent relay-chain blocks.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,342 @@
...
+
+This alludes to a simple pipeline where collators send validators parachain blocks and their requisite PoV to check. Then, validators validate the block using the PoV, signing statements which describe either the positive or negative outcome, and with enough positive statements, the block can be included. If another validator later detects that a validator or group of validators incorrectly signed a statement claiming a block was valid, then those validators will be _slashed_, with the checker receiving a bounty.
+
+However, there is a problem with this formulation. In order for another validator to check the previous group of validators' work after the fact, the PoV must remain _available_ so the other validator can fetch it in order to check the work. The PoVs are expected to be too large to include in the blockchain directly, so we require an alternate _data availability_ scheme which requires validators to prove that the inputs to their work will remain available, and so their work can be checked.
```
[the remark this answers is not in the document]
rphmeier: These PoVs will in practice be subsections of the state trie necessary to prove execution of a block. I did empirical measurements on Ethereum in 2016 which gave proof sizes of around 300Kb per block. We are using the same trie for Substrate chains (which, admittedly, is not the best trie for compact proofs), so we're expecting 10Mb to give us a ballpark estimate for how these might look under heavy load.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,342 @@
...
+
+This brings us to the second part of the process. Once a parablock is "fully included", it is still "pending approval". At this stage in the pipeline, the parablock has been backed by a majority of validators in the group assigned to that parachain, and its data has been guaranteed available by the set of validators as a whole. However, the validators in the parachain-group (known as the "Parachain Validators" for that parachain) are sampled from a validator set which contains some proportion of byzantine, or arbitrarily malicious members. This implies that the Parachain Validators for some parachain may be majority-dishonest, which means that secondary checks must be done on the block before it can be considered approved.
+
+The Approval Process looks like this:
```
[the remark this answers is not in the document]
rphmeier: Yeah, long story short is that you, over time, need to assign  >33% of your validators to check a specific parachain block. You can do that all at once in the first phase, but then you aren't sharding or scaling. The second phase (where we do secondary checks _after_ optimistic inclusion) gives us optimistic scaling with identical security properties.

I'll let @burdges, @FatemeShirazi, or @AlistairStewart add to that with any more details or correct me if I'm wrong.

This is definitely important background information, so I'll look into adding a short-and-sweet introduction to the sharding approach. Although, I don't think that we could put all the necessary background information into this doc - it would just be too much.

docs/polkadot/implementors-guide/guide.md
rphmeier: I'll add the link, but it doesn't seem like that glossary includes all of the terms I've been using, some of which are new. That's also a "Substrate" glossary, as opposed to a Polkadot one. cc @joshorndorff - I'll keep the terms here for now, but I'm hoping you can shed some light on the documentation story and where would be the best fit for terms that are mostly needed only to explain parachain-agreement internals.

docs/polkadot/implementors-guide/guide.md
rphmeier: I'd suggest to have a separate section which describes the Availability Distribution Process which then links to that PR. The initial description of the Availability Distribution Process will probably be a gossip-based one which only references @infinity0's work as future work.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
rphmeier: (I added the sharding intro - do you consider this resolved?)

rphmeier:
I pushed some changes writing up the general approach with processes and jobs for the node-side architecture. So far so good, but there are a couple points I think are worth discussing further:
  - Implicit dependencies between Processes. Each process will act as a server that receives a specific type of message, but we don't define which other processes are actually supposed to send those messages. This has the potential for clutter or difficult to follow logic.
  - Clunkiness of interprocess-communication. For basic functionality, the processes as laid out should work well, but more complex rules might become hard to encode and would require a lot of back-and-forth to avoid introducing a side-channel between processes. I was thinking of a specific scenario where we wanted more complex rules for how we choose the candidate we second based on the results of recent Candidate Backing Process jobs. I've been imagining that we'd have a separate process that talks to collators and finds candidates to second, and then sends them to the Candidate Backing Process to issue the statement. I don't see straight away how the communication of additional constraints from the backing process to the other process would work. This might say more about how we use the paradigm than the paradigm itself.

Previously when I thought about the Overseer, I imagined that there might be two overseers: one for the network, and one for the core. But now it seems better just to have the one. When I was thinking about the Overseer when writing the most recent sections, I was thinking we'd implement it in Rust as a big struct with a load of fields, one for each process. We're probably talking about the neighborhood of 15-30 processes. That should be simpler than trying to do something dynamic and reflective, especially when acting as a message bus - we can have a big enum with all the message variants and easily map each variant of that enum onto a specific field. Otherwise, if we go in the dynamic direction where we have the overseer just store something like a `Vec<AbstractProcessHandle>` we'd have to come up with a `Receiver`, `Filter`, `Listener` kind  of thing for knowing which kind of messages each type-abstracted process handle actually wants - sounds painful.

rphmeier:
> It would help to first write out the series of requirements of the runtime, so we can architect out how to carry out those tasks.

I've started on this now, but I don't know if I'll manage to push it tonight. This section is important and will also inform us about which core processes we need to write, based on what the runtime expects.

rphmeier:
Well, I managed to get through some of my thoughts on storage and the race conditions that are present, which should shed a lot of light on how we architect the storage and governance-managed configuration updates.

I didn't get through all of the things that the runtime has to do, but I will focus on those tomorrow and continue to work from there.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,500 @@
...
+* [Processes](#Processes)
+  * [Overseer](#Overseer-Process)
+  * [Candidate Backing](#Candidate-Backing)
+* [Data Structures and Types](#Data-Structures-and-Types)
+* [Glossary / Jargon](#Glossary)
+
+
+## Origins
+
+Parachains are the solution to a problem. As with any solution, it cannot be understood without first understanding the problem. So let's start by going over the issues faced by blockchain technology that let to us beginning to explore the design space for something like parachains.
```
drahnr, line 24: let -> led

rphmeier:
@mxinden and I had a discussion with these conclusions:
  - This is getting pretty close to an actor system (I believe that's a good thing)
  - We should clarify that the "processes" and "jobs" nomenclature is just a guideline meant to describe the way that we'd typically be using a hierarchical state machine architecture.
  - We shouldn't worry about restarting processes on failure - a process going down should take the whole system down.
  - It's better to avoid a monolithic networking process and instead split it up into many different processes. These processes should each manage their own (sub)protocol and should communicate with shared resources such as a peer-set manager via messages.
  - Processes that are tightly coupled (e.g. Candidate Backing & Statement Propagation) may be combined into their own local hierarchy to avoid the number of processes the overseer is directly exposed to and to make the coupling more clear. This should not prevent swapping out of implementations.
 - For the moment we're just considering gossiping everything, so we don't need to worry about more complex network protocols and how those would interact with sentry nodes and validator certificates yet. However, we should split the gossip protocols up so each process has its own `GossipEngine`. A network handle is required on startup but not afterwards.

(Sorry if I missed anything @mxinden )

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,500 @@
...
+
+There are three key points during the execution of a block that we are generally interested in:
+  * initialization: beginning the block and doing set up works. Runtime APIs draw information from the state directly after initialization.
+  * inclusion of new parachain information
+  * finalization: final checks and clean-up work before completing the block.
+
+In order to import parachains, handle misbehavior reports, and keep data accessible, we need to keep this data in the storage/state:
+  * All currently registered parachains.
+  * All currently registered parathreads.
+  * The head of each registered para.
```
pepyakin, line 153: What is "the head of a para"?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,500 @@
...
+
+This process receives messages of the type [CandidateBackingProcessMessage](#Candidate-Backing-Process-Message).
+
+#### Functionality
+
+The process should maintain a set of handles to Candidate Backing Jobs that are currently live, as well as the relay-parent to which they correspond.
+
+*On Overseer Signal*
+* If the signal is an `OverseerSignal::StartWork(relay_parent)`, spawn a Candidate Backing Job with the given relay parent, storing a bidirectional channel with the Candidate Backing Job in the set of handles.
+* If the signal is an `OverseerSignal::StopWork(relay_parent), cease the Candidate Backing Job under that relay parent, if any.
```
pepyakin, line 283:
```suggestion
* If the signal is an `OverseerSignal::StopWork(relay_parent)`, cease the Candidate Backing Job under that relay parent, if any.
```

docs/polkadot/implementors-guide/guide.md
rphmeier: We've used the shorthand "para" to mean "parachain or parathread", but I don't know if I've defined this anywhere.

pepyakin: Actually my question was more about the "head" part : ) It seems that it is not defined anywhere around

rphmeier: oh, by that I meant the current head-data.

Everything I'm writing in this section right now is super messy because I'm just trying to put down all the requirements of the runtime, have others look over them, and then start to propose an architecture that encompasses all of them.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,500 @@
...
... the second part of the process. Once a parablock is "fully included", it is still "pending approval". At this stage in the pipeline, the parablock has been backed by a majority of validators in the group assigned to that parachain, and its data has been guaranteed available by the set of validators as a whole. However, the validators in the parachain-group (known as the "Parachain Validators" for that parachain) are sampled from a validator set which contains some proportion of byzantine, or arbitrarily malicious members. This implies that the Parachain Validators for some parachain may be majority-dishonest, which means that secondary checks must be done on the block before it can be considered approved. This is necessary only because the Parachain Validators for a given parachain are sampled from an overall validator set which is assumed to be up to <1/3 dishonest - meaning that there is a chance to randomly sample Parachain Validators for a parachain that are majority or fully dishonest and can back a candidate wrongly. The Approval Process allows us to detect such misbehavior after-the-fact without allocating more Parachain Validators and reducing the throughput of the system.
```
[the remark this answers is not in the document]
rphmeier: I've reworded this locally to avoid this phrasing - will push later tonight.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
rphmeier: I added a note on the empirical size

docs/polkadot/implementors-guide/guide.md
```diff
@@ -25,7 +25,17 @@ Parachains are the solution to a problem. As with any solution, it cannot be und
...
+
+Proof-of-Stake (PoS) systems can accomplish higher throughput than PoW blockchains. PoS systems are secured by bonded capital as opposed to spent effort - liquidity opportunity cost vs. burning electricity. The way they work is by selecting a set of validators with known economic identity who lock up tokens in exchange for earning the right to "validate" or participate in the consensus process. If they are found to carry out that process wrongly, their tokens will be burned. This provides a strong disincentive in the direction of misbehavior.
+
+Since the consensus protocol doesn't revolve around wasting effort, block times and agreement can occur much faster. Solutions to PoW challenges don't have to be found before a block can be authored, so the overhead of authoring a block is reduced to only the costs of creating and distributing the block.
+
+However, consensus on a PoS chain requires full agreement of 2/3+ of the validator set for everything that occurs at Layer 1: all logic which is carried out as part of the blockchain's state machine. This means that everybody still needs to check everything.
```
FatemeShirazi, line 34: I would add here that communication among large number validators with gossiping means that data arrives at validators with different speed and the view of the systems is not synchronized. Some might be a couple of blocks behind others in what they have received. This makes the agreement among them more complicated.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -25,7 +25,17 @@ Parachains are the solution to a problem. As with any solution, it cannot be und
 
 #### Issue 1: Scalability
 
-It became clear a few years ago that the transaction throughput of simple Proof-of-Work blockchains such as Bitcoin, Ethereum, and myriad others was simply too low. [TODO: PoS, sharding, what if there were more blockchains, etc. etc.]
+It became clear a few years ago that the transaction throughput of simple Proof-of-Work (PoW) blockchains such as Bitcoin, Ethereum, and myriad others was simply too low. [TODO: PoS, sharding, what if there were more blockchains, etc. etc.]
+
+Proof-of-Stake (PoS) systems can accomplish higher throughput than PoW blockchains. PoS systems are secured by bonded capital as opposed to spent effort - liquidity opportunity cost vs. burning electricity. The way they work is by selecting a set of validators with known economic identity who lock up tokens in exchange for earning the right to "validate" or participate in the consensus process. If they are found to carry out that process wrongly, their tokens will be burned. This provides a strong disincentive in the direction of misbehavior.
+
```
FatemeShirazi, line 31: ...will be burned (or slashed). This ...

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
FatemeShirazi: Even with random assignment in order to have only one pipeline, the number of parachain validators assigned to a parachain needs to be quite large to make sure for every parachain majority of assigned parachain validators is honest. As Rob mentioned, having two pipelines allows us to get away with scaling very well in case validators are honest and having an assurance that if that is not the case it will be caught by the second pipeline. The second pipeline has to happen anyway also, once we have any fishermen reporting faulty behavior. In fact, the more reports we have from fishermen the number of secondary checkers increases.

docs/polkadot/implementors-guide/guide.md
infinity0: single finalised head, or all the unfinalised heads plural?

infinity0: @pepyakin head refers to the tip (most-recent) nodes of a DAG  / history data structure like HEAD in git

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+  Seconded(CandidateReceipt),
+  /// A statement about the validity of a candidate, based on candidate's hash.
+  Valid(Hash),
+  /// A statement about the invalidity of a candidate.
+  Invalid(Hash),
+}
+```
+
+#### Signed Statement Type
+
```
drahnr, line 462: `This is a signed statement.` does not add much value to the block :)

Not sure if you want to go into the purpose in this section.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+  statement: Statement,
+  signed: ValidatorId,
+  signature: Signature
+}
+```
+
+
+#### Overseer Signal
+
+Signals from the overseer to a process.
```
drahnr, line 479: `...to request change in execution that has to be obeyed by the process.` to clarify things a bit more based on the previous sections, maybe too verbose.

docs/polkadot/implementors-guide/guide.md
JoshOrndorff: The Substrate now lives here https://www.substrate.io/kb/getting-started/glossary and the source is here https://github.com/substrate-developer-hub/knowledgebase/blob/master/current/getting-started/glossary.md

I agree we want to keep it a Substrate glossary and not introduce polkadot-specific terms. There is also the polkadot wiki's glossary which probably makes more sense in this case https://wiki.polkadot.network/docs/en/glossary

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+
+Having recognized these issues, we set out to find a solution to these problems, which could allow developers to create and deploy purpose-built blockchains unified under a common source of security, with the capability of message-passing between them; a _heterogeneous sharding solution_, which we have come to know as **Parachains**.
+
+
+----
+
+## Parachains: Basic Functionality
+
+This section aims to describe, at a high level, the architecture, actors, and processes involved in the implementation of parachains. It also illuminates certain subtleties and challenges faces in the design and implementation of those processes. Our goal is to carry a parachain block from authoring to secure inclusion, and define a process which can be carried out repeatedly and in parallel for many different parachains to extend them over time. Understanding of the high-level approach taken here is important to provide context for the proposed architecture further on.
```
JoshOrndorff, line 54:
```suggestion
This section aims to describe, at a high level, the architecture, actors, and processes involved in the implementation of parachains. It also illuminates certain subtleties and challenges faced in the design and implementation of those processes. Our goal is to carry a parachain block from authoring to secure inclusion, and define a process which can be carried out repeatedly and in parallel for many different parachains to extend them over time. Understanding of the high-level approach taken here is important to provide context for the proposed architecture further on.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+First, it's important to go over the main actors we have involved in the Polkadot network.
+1. Validators. These nodes are responsible for validating proposed parachain blocks. They do so by checking a Proof-of-Validity (PoV) of the block and ensuring that the PoV remains available. They put financial capital down as "skin in the game" which can be revoked if they are proven to have misvalidated.
```
JoshOrndorff, line 57: Use of "revoked" here made me think the validator could choose to take their skin out of the game and no longer support a PoV block that they previously supported. You previously used "slashed" or "burned" which I think are better.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+2. Collators. These nodes are responsible for creating the Proofs-of-Validity that validators know how to check. Creating a PoV typically requires familiarity with the transaction format and block authoring rules of the parachain, as well as having access to the full state of the parachain.
+3. Fishermen. These are user-operated, permissionless nodes whose goal is to catch out misbehaving validators in exchange for a bounty. Collators and validators can behave as Fishermen too. Fishermen aren't necessary for security, and aren't covered in-depth by this document.
```
JoshOrndorff, line 59:
```suggestion
3. Fishermen. These are user-operated, permissionless nodes whose goal is to catch misbehaving validators in exchange for a bounty. Collators and validators can behave as Fishermen too. Fishermen aren't necessary for security, and aren't covered in-depth by this document.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+However, there is a problem with this formulation. In order for another validator to check the previous group of validators' work after the fact, the PoV must remain _available_ so the other validator can fetch it in order to check the work. The PoVs are expected to be too large to include in the blockchain directly, so we require an alternate _data availability_ scheme which requires validators to prove that the inputs to their work will remain available, and so their work can be checked. Empirical tests tell us that many PoVs may be between 1 and 10MB during periods of heavy load.
+
+Here is a description of the Inclusion Pipeline: the path a parachain block (or parablock, for short) takes from creation to inclusion:
+1. Validators are selected and assigned to parachains by the Validator Assignment Process.
```
JoshOrndorff, line 66:
```suggestion
1. Validators are selected and assigned to parachains by the Validator Assignment Process, a process that the parachain host executes on-chain.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+1. A collator produces the parachain block, which is known as a parachain candidate or candidate, along with a PoV for the candidate.
+1. The collator forwards the candidate and PoV to validators assigned to the same parachain via the Collation Distribution Process.
+1. The validators assigned to a parachain at a given point in time participate in the Candidate Backing Process to validate candidates that were put forward for validation. Candidates which gather enough signed validity statements from validators are considered "backed" and are called backed candidates. Their backing is the set of signed validity statements.
+1. A relay-chain block author, selected by BABE, can include up to one (1) backed candidate for each parachain to include in the relay-chain block alongside its backing.
```
JoshOrndorff, line 70: You've switched from "parachain host" the "relay chain" here. Not the worst thing, but thought I'd share the observation.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+The Approval Process looks like this:
+1. Parablocks that have been included by the Inclusion Pipeline are pending approval for a time-window known as the secondary checking window.
+1. During the secondary-checking window, validators randomly self-select to perform secondary checks on the parablock.
```
JoshOrndorff, line 87: Does "randomly self-select" mean they do it whenever they feel like it? Or they use some verifiable random process? Is there any incentive or mandate to self-select?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+It is also important to take note of the fact that the relay-chain is extended by BABE, which is a forkful algorithm. That means that different block authors can be chosen at the same time, and may not be building on the same block parent. Furthermore, the set of validators is not fixed, nor is the set of parachains. And even with the same set of validators and parachains, the validators' assignments to parachains is flexible. This means that the architecture proposed in the next chapters must deal with the variability and multiplicity of the network state.
+
+[TODO Diagram: Forkfulness]
+
+----
+
+## Architecture
+
+Our Parachain Host is a blockchain. A blockchain is a Directed Acyclic Graph (DAG) of state transitions, where every block can be considered to be the head of a linked-list (known as a "chain" or "fork") with a cumulative state which is determined by applying the state transition of each block in turn. All paths through the DAG terminate at the Genesis Block.
```
JoshOrndorff, line 104: "A blockchain is a Directed Acyclic Graph (DAG) of state transitions"

That is correct, of course, but a blockchain is also a tree, a more specific subset of DAGs. I think calling it a DAG here will give the impression that fork merging is possible which is not the case for the parachain host.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+[TODO Diagram: Node-side divided into Networking and Core]
+
+Node-side behavior is split up into various Processes. Processes are long-lived workers that perform a particular category of work. Processes can communicate with each other, and typically do so via an Overseer that prevents race conditions.
+
+Runtime logic is divided up into Modules and APIs. Modules encapsulate particular behavior of the system. Modules consist of storage, routines, and entry-points. Routines are invoked by entry points, by other modules, upon block initialization or closing. Routines can read and alter the storage of the module. Entry-points are the means by which new information is introduced to a module. Each block in the blockchain contains a set of Extrinsics. Each extrinsic targets a a specific entry point to trigger and which data should be passed to it. Runtime APIs provide a means for Node-side behavior to extract meaningful information from the state of a fork.
```
JoshOrndorff, line 126:
```suggestion
Runtime logic is divided up into Modules and APIs. Modules encapsulate particular behavior of the system. Modules consist of storage, routines, and entry-points. Routines are invoked by entry points, by other modules, upon block initialization or closing. Routines can read and alter the storage of the module. Entry-points are the means by which new information is introduced to a module. Each block in the blockchain contains a set of Extrinsics. Each extrinsic targets a a specific entry point to trigger and which data should be passed to it. Runtime APIs provide a means for Node-side behavior to extract meaningful information from the state of a single fork.
```

docs/polkadot/implementors-guide/guide.md
rphmeier: Yeah, they use a verifiable random process. I assume they are meant to be rewarded for it, but @burdges can shed more light on that.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+
+
+## Origins
+
+Parachains are the solution to a problem. As with any solution, it cannot be understood without first understanding the problem. So let's start by going over the issues faced by blockchain technology that let to us beginning to explore the design space for something like parachains.
+
+#### Issue 1: Scalability
+
+It became clear a few years ago that the transaction throughput of simple Proof-of-Work (PoW) blockchains such as Bitcoin, Ethereum, and myriad others was simply too low. [TODO: PoS, sharding, what if there were more blockchains, etc. etc.]
+
```
montekki, line 29: A bit on security also:

```suggestion

The other concern is the vulnerability of newly created PoW blockchains to  51% attacks. An amount of existing mining power in the world is finite and it is unlikely that a newly created blockchain would attract as much mining power as the existing popular PoW blockchains. This fact makes 51% attacks against our newly created blockchain relatively cheap. The estimated cost of 1 hour attack on Bitcoin is $700000/hour however for a a huge number of less popular PoW blockchains the estimate cost of an attack under $100/hour (at the time of writing of this document). It is obvious that implementing a new blockchain with a PoW consensus is orthogonal to creating a system with a reasonable level of security.

```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+  * finalization: final checks and clean-up work before completing the block.
+
+In order to import parachains, handle misbehavior reports, and keep data accessible, we need to keep this data in the storage/state:
+  * All currently registered parachains.
+  * All currently registered parathreads.
+  * The head of each registered parachain or parathread.
+  * The validation code of each registered parachain or parathread.
+  * Historical validation code for each registered parachain or parathread.
+  * Historical, but not yet expired validation code for paras that were previously registered but are now not. (old code must remain available so secondary checkers can check after-the-fact yadda yadda in this case we do that by keeping it in the runtime state.)
+  * Configuration: number of parathread cores, number of parachain slots. Length of scheduled parathread "lookahead". Length of parachain slashing period. How long to keep old validation code for. etc.
```
montekki, line 162:
```suggestion
  * Configuration: number of parathread cores, number of parachain slots. Length of scheduled parathread "lookahead". Length of parachain slashing period. How long to keep old validation code for. etc.
  * Historical data for validators sets at least [TODO: how many?] blocks into the past. Used when reporting equivocations to prove that the validator at question actually belonged to the validator set at the time the equivocation was commited.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,544 @@
...
+  * Schedule Upwards messages - messages from the parachain to the relay chain.
+
+process new backed candidates:
+  * ensure that only one candidate is backed for each parachain or parathread
+  * ensure that the parachain or parathread of the candidate was scheduled and does not currently have a block pending availability. 
+  * check the backing of the candidate. 
+  * move to "pending approval" state. (pass along any configuration information that is liable to change)
+
+misbehavior reports and secondary checks:
+  * Secondary checks will also be submitted within the block. This may lead to slashing as a secondary check period ends.
```
montekki, line 204:
```suggestion
  * Secondary checks will also be submitted within the block. This may lead to slashing as a secondary check period ends. We want to catch and punish for the cases of misbehavior that violate the protocol and put it's security at risk. One of such cases is submitting conflicting votes on the same `CandidateReceipt`.Other examples include violations to AnV protocol or equivocations in finality. Misbehavior handling is implemented in
    * Runtime as an entry point.
    * Code in the Node that assists submitting misbehavior reports.
```

docs/polkadot/implementors-guide/guide.md
rphmeier: single finalized head for the view of each relay chain block's state. many unfinalized relay chain heads -> many unfinalized parachain heads.

docs/polkadot/implementors-guide/guide.md
rphmeier: This may be too much information for this preliminary section - I'm not sure it's relevant whether the validator assignments are determined on or off-chain at this overview. When the section on validator assignment is completed, we would link to that section. What do you think?

docs/polkadot/implementors-guide/guide.md
rphmeier: I think it's unavoidable to talk about the relay chain if we want to get into sufficient detail here, so i'll specify that a parachain host _has_ a blockchain known as a relay chain

docs/polkadot/implementors-guide/guide.md
rphmeier: The statement that each block is the head of a linked-list should technically does this already, but I will clarify.

docs/polkadot/implementors-guide/guide.md
rphmeier: Do you have a source on that information? It may not fit in the flow of a "scalability" section, but there could be another section on PoS vs PoW.

montekki: > Recently there have been a number of 51% attacks including a high profile attack against Bitcoin Gold where $18 Million was stolen.

https://www.crypto51.app/

docs/polkadot/implementors-guide/guide.md
rphmeier: Who's responsible for the Polkadot glossary? Is that on Web3's side?

JoshOrndorff: On Web3's side. I'm not sure who's responsible.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
AlistairStewart: >long story short is that you, over time, need to assign >33% of your validators to check a specific parachain block.

This is not true as written. Without challenges, only a few validators check a specific parachain block.

The goal here is to make attacking Polkadot expensive, while scaling as much as possible. We can get stronger security with fewer checkers with this split into backing and approval checkers as Fatemeh explained.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -63,7 +63,22 @@ Note that the candidate can fail to be included in any of the following ways:
...
+1. At the end of the Approval Process, the parablock is either Approved or it is rejected. More on the rejection process later.
+
+These two pipelines sum up the sequence of events necessary to extend and acquire full security on a Parablock. Note that the Inclusion Pipeline must conclude for a specific parachain before a new block can be accepted on that parachain. After inclusion, the Approval Process kicks off, and can be running for many parachain blocks at once.
+
+[TODO Diagram: Inclusion Pipeline & Approval Processes interaction]
+
+It is also important to take note of the fact that the relay-chain is extended by BABE, which is a forkful algorithm. That means that different block authors can be chosen at the same time, and may not be building on the same block parent. Furthermore, the set of validators is not fixed, nor is the set of parachains. And even with the same set of validators and parachains, the validators' assignments to parachains is flexible. This means that the architecture proposed in the next chapters must deal with the variability and multiplicity of the network state.
```
AlistairStewart, line 98: It might go without saying that while the code does need to deal with all this, an important design goal is to minimize disagreement between forks on things like parachain assignments.

AlistairStewart:
I can't see much in this document that I'd argue with, and it will be very helpful to the research team as well as devs.

docs/polkadot/implementors-guide/guide.md
JoshOrndorff: Sure, that's fine.

expenses:
When we say a job is 'spawned', we're could either be talking about a spawned future or a thread. Futures are probably the better choice, but this should be clarified.

I'm less sure about which futures executor we want. As far as I know, using `async-std` would force us into a framework less than using `tokio`. @mxinden do you know much about this?

In order to cancel futures upon a `OverseerSignal::StopWork` call, we need to hold their handles somewhere, along with the relay chain block that they're running for. If possible, it would be more modular to store these separately for each process.

docs/polkadot/implementors-guide/guide.md
burdges: Yes validators use VRFs for this.

If validators self selected their secondary approval checker assignments using only local non-verifiable randomness then well-funded but "overly curious" validators could download all parachains blocks across the network, which (a) reduces our scalability and worse (b) reduces our tricks against adaptive DoSing adversaries.

Instead, all validators have some VRF key with which they compute VRF values that determine their assignments.  If V announces a VRF that says V checks B then (a) V has highest priority for receiving pieces of B, but also (b) we do not proceed into grandpa until either V checks B, or else we declare V a no-show and solicit more previously unknown VRFs.

A strong DoS adversary could force multiple nodes be no-shows until their own nodes fulfilled all required checks.  We expect they'd need to silence a couples times the expected number of checkers though, which increases the odds the attacker hits some hardened honest node with many and/or hidden sentries.  

We hope honest no-shows would be rare because if you're validator was online to gossip out a VRF then they should be there to fetch pieces immediately afterwards, and thus hopefully making them a strong sign of DoS attacks.

Initially we'll need to activate parachains without all this no-show logic because selecting the delay requires real live network experience but that's fine.

mxinden:
> I'm less sure about which futures executor we want. As far as I know, using async-std would force us into a framework less than using tokio. @mxinden do you know much about this?

Why not use whatever executor coming with Substrate and spawn tasks onto it?

expenses:
> Why not use whatever executor coming with Substrate and spawn tasks onto it?

Oh yes, that works as well.

rphmeier:
> In order to cancel futures upon a OverseerSignal::StopWork call, we need to hold their handles somewhere, along with the relay chain block that they're running for. If possible, it would be more modular to store these separately for each process.

Agreed, better for processes to track anything they spawn.

I have been writing this with Futures in mind as opposed to threads but I haven't given much thought to executor. Substrate's executor, which is `tokio` AFAIK should be fine

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,660 @@
...
+misbehavior reports and secondary checks:
+  * Secondary checks will also be submitted within the block. This may lead to slashing as a secondary check period ends. We want to catch and punish for the cases of misbehavior that violate the protocol and put its security at risk. One of such cases is submitting conflicting votes on the same `CandidateReceipt`.Other examples include violations to AnV protocol or equivocations in finality. Misbehavior handling is implemented in
+    * Runtime as an entry point.
+    * Code in the Node that assists submitting misbehavior reports.
+
+finalization: (not finality)
+  * ensure that required updates (bitfields and backed candidates) occurred within the block.
+  * update scheduling metadata based on parachains that had blocks included or not. for instance, parathreads where the auction-winning collator didn't get a chance to include its block should be allowed to retry a couple of times.
+
+Availability bitfields must go in before parachain candidates, otherwise there would be a minimum of 1 relay chain block between blocks of the same parachain. As such, it's best for them to go into the same extrinsic.
```
burdges, line 326: We could restrict the block structure elsewhere too instead, right?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,660 @@
...
+
+### Candidate Backing Process
+
+#### Description
+
+The Candidate Backing Process is the process that a validator engages in to contribute to the backing of parachain candidates submitted by other validators.
+
+Its role is to produce backed candidates for inclusion in new relay-chain blocks. It does so by issuing signed [Statements](#Statement-type) and tracking received statements signed by other validators. Once enough statements are received, they can be combined into backing for specific candidates.
+
+It also detects double-vote misbehavior by validators as it imports votes, passing on the misbehavior to the correct reporter and handler.
```
burdges, line 429: You mean double initial proposal of a candidate?  We want parachain backing validators to votes for one anothers' suggestions, but the block producer makes the choice.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+----
+
+## Processes
+
+### Processes and Jobs
+
+In this section we define the notions of Processes and Jobs. These are guidelines for how we will employ an architecture of hierarchical state machines. We'll have a top-level state machine which oversees the next level of state machines which oversee another layer of state machines and so on. The next sections will lay out these guidelines for what we've called Processes and Jobs, since this model applies to many of the tasks that the Node-side behavior needs to encompass, but these are only guidelines and some processes may have deeper hierarchies internally.
+
+Processes are long-lived worker tasks that are in charge of performing some particular kind of work. All processes can communicate with each other via a well-defined protocol. Processes can't communicate directly, but must communicate through an Overseer, which is responsible for relaying messages, handling process failures, and dispatching work signals.
```
coriolinus, line 502: Do all processes have direct access to the Overseer to communicate, or do they always pass messages strictly up and down through the hierarchy?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+  header: Header,
+}
+```
+
+#### Statement Type
+```rust
+/// A statement about the validity of a parachain candidate.
+enum Statement {
+  /// A statement about a new candidate being seconded by a validator. This is an implicit validity vote.
+  Seconded(CandidateReceipt),
```
coriolinus, line 732:
```suggestion
  Seconded(Hash),
```

Per the first line of the  "Signed Statement Type" description.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+         |                   |<--  from "bar" --|                    |
+         |                   |                  |                    |
+         \___________________/                  \____________________/
+                                                   ______| |______
+                                                   ___Transport___
+
+```                                                   
+
+
+Node-side behavior is split up into various Processes. Processes are long-lived workers that perform a particular category of work. Processes can communicate with each other, and typically do so via an Overseer that prevents race conditions.
```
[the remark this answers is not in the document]
expenses: Yeah, this wording is strange to me as well. Perhaps calling them something like 'Systems' or 'Subsystems' would work better?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+Here is an attempted-exhaustive list of tasks the runtime is expected to carry out in each phase.
+
+initialization:
+  * accept new registrations of parachains and parathreads. Probably best to do this only once a session to avoid bitfield schemas shifting often (see details on availability bitfields below)
```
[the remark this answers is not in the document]
expenses: https://wiki.polkadot.network/docs/en/glossary#session

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+#### Overseer Signal
+
+Signals from the overseer to a process to request change in execution that has to be obeyed by the process.
+
+```rust
+enum OverseerSignal {
+  /// Signal to start work localized to the relay-parent hash H.
+  StartWork(H),
+  /// Signal to stop (or phase down) work localized to the relay-parent hash H.
+  StopWork(H),
```
expenses, line 765: Might as well be consistent:

```suggestion
  /// Signal to start work localized to the relay-parent hash.
  StartWork(Hash),
  /// Signal to stop (or phase down) work localized to the relay-parent hash.
  StopWork(Hash),
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+#### Candidate Backing Process Message
+
+```rust
+enum CandidateBackingProcessMessage {
+  /// Registers a stream listener for updates to the set of backed candidates that could be included
+  /// in a child of the given relay-parent, referenced by its hash H
+  RegisterBackingWatcher(H, TODO),
+  /// Note that the Candidate Backing Process should second the given candidate in the context of the 
+  /// given relay-parent (ref. by hash H). This candidate must be validated.
+  Second(H, CandidateReceipt)
```
expenses, line 779:
```suggestion
  /// given relay-parent (ref. by hash). This candidate must be validated.
  Second(Hash, CandidateReceipt)
```

infinity0:
whoops, closed this by mistake & have undone. I just added an index page so it now gets built and published onto the website. Also I'd suggest the README.md is redundant and should just be dropped. Also please rebase onto master

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+Parachains are an example of a **sharded** protocol. Sharding is a concept borrowed from traditional database architecture. Rather than requiring every participant to check every transaction, we require each participant to check some subset of transactions, with enough redundancy baked in that byzantine (arbitrarily malicious) participants can't sneak in invalid transactions - at least not without being detected and getting slashed, with those transactions reverted.
+
+Sharding and Proof-of-Stake in coordination with each other allow a parachain host to provide full security on many parachains, even without all participants checking all state transitions.
+
+[TODO: note about network effects & bridging]
+
+#### Issue 2: Flexibility / Specialization
+
+"dumb" VMs don't give you the flexibility. Any engineer knows that being able to specialize on a problem and create coarser-grained operations gives them and their users more _leverage_.  [TODO]
```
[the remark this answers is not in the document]
rphmeier: I meant it in the sense of coarse vs. fine-grained sand, but agreed that the metaphor is distracting.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+The Parachain Host is a blockchain, known as the relay-chain, and the actors which provide security and inputs to the blockchain.
+
+First, it's important to go over the main actors we have involved in the parachain host.
+1. Validators. These nodes are responsible for validating proposed parachain blocks. They do so by checking a Proof-of-Validity (PoV) of the block and ensuring that the PoV remains available. They put financial capital down as "skin in the game" which can be slashed, or burned if they are proven to have misvalidated.
+2. Collators. These nodes are responsible for creating the Proofs-of-Validity that validators know how to check. Creating a PoV typically requires familiarity with the transaction format and block authoring rules of the parachain, as well as having access to the full state of the parachain.
+3. Fishermen. These are user-operated, permissionless nodes whose goal is to catch misbehaving validators in exchange for a bounty. Collators and validators can behave as Fishermen too. Fishermen aren't necessary for security, and aren't covered in-depth by this document.
```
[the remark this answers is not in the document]
rphmeier: cc @FatemeShirazi which document is best to link to for this property? Jeff's AnV paper or the overview paper?

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+This alludes to a simple pipeline where collators send validators parachain blocks and their requisite PoV to check. Then, validators validate the block using the PoV, signing statements which describe either the positive or negative outcome, and with enough positive statements, the block can be included. If another validator later detects that a validator or group of validators incorrectly signed a statement claiming a block was valid, then those validators will be _slashed_, with the checker receiving a bounty.
```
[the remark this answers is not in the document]
rphmeier: At this level of abstraction, I don't believe we have to address the semantics of what should be considered a part of the parachain or not. There are good arguments in either direction, but none should affect what the parachain host is doing.

> If a validator incorrectly signs a statement that a block was invalid, is that validator slashed?

Yes. I haven't been writing all that much AnV stuff in, but being wrong about validity or invalidity is a 100% slash.

I'll make some amendments to clarify these points

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+1. A relay-chain block author, selected by BABE, can include up to one (1) backed candidate for each parachain to include in the relay-chain block alongside its backing.
+1. Once included in the relay-chain, the parachain candidate is considered to be "pending availability". It is not considered to be part of the parachain until it is proven available.
+1. In the following relay-chain blocks, validators will participate in the Availability Distribution Process to ensure availability of the candidate. Information regarding the availability of the candidate will be included in the subsequent relay-chain blocks.
+1. Once the relay-chain state machine has enough information to consider the candidate's PoV as being available, the candidate is considered to be part of the parachain and is graduated to being a full parachain block, or parablock for short.
+
+Note that the candidate can fail to be included in any of the following ways:
+  - The collator is not able to propagate the candidate to any validators assigned to the parachain.
+  - The candidate is not fully backed by validators participating in the Candidate Backing Process.
```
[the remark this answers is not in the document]
rphmeier: This should just be "backed" - I had used the terminology "fully backed" in an earlier draft and changed it everywhere but here.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+The Approval Process looks like this:
+1. Parablocks that have been included by the Inclusion Pipeline are pending approval for a time-window known as the secondary checking window.
+1. During the secondary-checking window, validators randomly self-select to perform secondary checks on the parablock.
+1. These validators, known in this context as secondary checkers, acquire the parablock and its PoV, and re-run the validation function.
+1. The secondary checkers submit the result of their checks to the relay chain. Contradictory results lead to escalation, where even more secondary checkers are selected and the secondary-checking window is extended.
+1. At the end of the Approval Process, the parablock is either Approved or it is rejected. More on the rejection process later.
```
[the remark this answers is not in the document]
rphmeier: No, a parablock is a valid parent as soon as it becomes available. Collators will even want to act optimistically on the assumption that a block will become available before it has, to maximize throughput of the parachain.

The collator needs to watch for new relay-chain blocks and observe if any new blocks have become available (or are pending availability). We will need to expose runtime APIs or stable storage locations for collators to watch that.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+                      +----------------------+    
+                      |  Relay Block B       |
+                      +----------------------+
+	                             |
+                      +----------------------+
+                      |  Relay Block A       |
+                      +----------------------+
+```
+
+Those validators that are aware of many competing heads must be aware of the work happening on each one. They may contribute to some or a full extent on both. It is possible that due to network asynchrony two forks may grow in parallel for some time, although in the absence of an adversarial network this is unlikely in the case where there are validators who are aware of both chain heads.
```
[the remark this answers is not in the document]
rphmeier: This falls more into the domain of our block production system BABE (or soon, SASSAFRAS). First, those validators who are aware of both chain heads should be propagating them onwards to all of their peers of the network. The block-author selection algorithm and incentives are designed to ensure that only one longest chain will prevail as long as there is no network partition.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
rphmeier: Let's rename them then because I've had the same feeling that there's too much potential confusion between Process (parachain host) and Process (OS). 
 
I believe @montekki was advocating for "Worker" in the past. I'm not strongly opinionated on either that or Subsystem

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
rphmeier: This section is mostly collaborative scratch and will disappear into the runtime module architecture proposal below. I believe I go into more details on sessions there.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+
+Here is an attempted-exhaustive list of tasks the runtime is expected to carry out in each phase.
+
+initialization:
+  * accept new registrations of parachains and parathreads. Probably best to do this only once a session to avoid bitfield schemas shifting often (see details on availability bitfields below)
+  * determine scheduled parachains and parathreads for the upcoming block or blocks.
+  * determine validator assignments to scheduled paras for the upcoming block or blocks.
+  * remove blocks which have been pending availability for too long. this is tightly coupled with scheduling.
+  * handle the start of a new session - discard all candidates pending availability and note the upcoming validator set.
```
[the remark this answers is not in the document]
rphmeier: This is something else. We want to have a timeout of a certain amount of blocks for availability to be reached. But for consistency's sake it's really hard to handle blocks that are pending availability across session boundaries, so we just cancel all of them at the beginning of every session. It's a simple solution that makes the model more clean, but I'm hoping that a later draft of this document will be less draconian.

docs/polkadot/implementors-guide/guide.md
rphmeier: Processes exist one level down from the overseer in the hierarchy of asynchronous state machines. They will have a bidirectional channel with the overseer for communicating with it.

docs/polkadot/implementors-guide/guide.md
rphmeier: No, we actually want the `CandidateReceipt` to be included in full here. It's just that the payload that's actually signed in the `SignedStatement` shouldn't encode the full candidate receipt, but only its hash. That allows us to omit the receipt elsewhere. Do you have any ideas on how to communicate that better?

The reason we want to include the `CandidateReceipt` in the `Statement::Seconded` is so we can ensure we have all relevant `CandidateReceipt`s in a data structure which only stores `Statement`s. Otherwise you'd always have to carry along an `Option<CandidateReceipt>` which would only be `Some` if the `Statement` was `Seconded`, as seconding is the introductory point of a candidate receipt to the import pipeline.

rphmeier:
Renamed Process -> Subsystem as it avoid confusion with OS processes and other uses of the verb "to process".

docs/polkadot/implementors-guide/guide.md
rphmeier: Missed this until now. could you elaborate on this comment? I'm not sure what you mean by it

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,1013 @@
...
+         |                   |<--  from "bar" --|                    |
+         |                   |                  |                    |
+         \___________________/                  \____________________/
+                                                   ______| |______
+                                                   ___Transport___
+
+```                                                   
+
+
+Node-side behavior is split up into various subsystems. Subsystems are long-lived workers that perform a particular category of work. Subsystems can communicate with each other, and do so via an Overseer that prevents race conditions.
```
mxinden, line 241: > that prevents race conditions

For completeness sake, would you mind describing how the serialization point *Overseer* helps prevent race conditions?

docs/polkadot/implementors-guide/guide.md
burdges: We could enforce extrinsics appear in some specified order inside the block, no?  

We read some series of `CandidateBacked`, `AvailabilityVote`, and `CandidateAvailable` extrinsics.  I think `CandidateAvailable` refers to a specific `CandidateReceipt` but otherwise only states that enough `AvailabilityVote`s happened to trigger tallies during block verification.  We cannot avoid `AvailabilityVote`s each having their own relay chain parent, possibly the chain head, but not the current block.  We mask out any new `CandidateBacked` transactions between the votes parent and the chain head when processing, so if we process any `CandidateBacked` before a corresponding `CandidateAvailable` then we remove that bit form the mask, and block that candidate being available.  It sounds okay, although might enable censorship of parathreads, so we could invalidate the block if any `AvailabilityVote`s or `CandidateAvailable` have parents that precede the current `CandidateBacked`.  In effect, this imposes the desired block structure.  

Any threads processing `AvailabilityVote`s for the same candidate must be joined before its `CandidateAvailable` I guess.  I think `CandidateBacked` could safely run inside another thread even under the weaker memory model.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,854 @@
...
+----
+
+## Parachains: Basic Functionality
+
+This section aims to describe, at a high level, the architecture, actors, and processes involved in the implementation of parachains. It also illuminates certain subtleties and challenges faced in the design and implementation of those processes. Our goal is to carry a parachain block from authoring to secure inclusion, and define a process which can be carried out repeatedly and in parallel for many different parachains to extend them over time. Understanding of the high-level approach taken here is important to provide context for the proposed architecture further on.
+
+The Parachain Host is a blockchain, known as the relay-chain, and the actors which provide security and inputs to the blockchain.
+
+First, it's important to go over the main actors we have involved in the parachain host.
+1. Validators. These nodes are responsible for validating proposed parachain blocks. They do so by checking a Proof-of-Validity (PoV) of the block and ensuring that the PoV remains available. They put financial capital down as "skin in the game" which can be slashed, or burned if they are proven to have misvalidated.
```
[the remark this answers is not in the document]
burdges: A PoV block is a parachain block along with witnesses that enable stateless verification, which currently consist of Merkle proofs into the chains' state.  If the block has size n ad the state has size m then then the PoV has size roughly n log m.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
burdges: We argue fishermen are unnecessary in the sense that our availability system is optimized to increase the power of honest validators, who then deliver confidence and finality quickly.  

We do think fishermen improve longer term security somehow.  In particular they could increase confidence among a wider audience.   We cannot make claims about fishermen rigorous without designing an availability system for fishermen however, which is what ETH2, etc. do, but gets very slow.  

We dislike some assumptions underlying fishermen centric designs too, seemingly including non-adaptive adversaries, which again if you assume away DoS attacks when you make the system establish security slower. 

We put time into mixed availability systems that serve both validators and fishermen, but they'd incur significant costs. 

tl;dr Actually fishermen do improve security but we donno how much, when, how quickly, etc.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
burdges: We could permit less than 100% slash for being wrong about invalidity, but it must remain quite significant so that validator do not just vote invalidity to get others slashed, and I do not see any reason why it needs to be less than 100%.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
burdges: Actually one backing validator would suffice, but more seems better.  50% maybe?

Any absent backer below 100% of the parachain validators gets replaced with an extra (secondary) approval checkers.  I'd consider that overkill, well every approval checker should be worth more than one parachain validators, but no idea how many.  1.5?  2?  5?  20?

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
burdges: Yes.  Approval could be recorded on-chain, but it's really there to limit finality in grandpa.  A bridge or outside service should normally wait upon grandpa and thus wait on approval.

docs/polkadot/implementors-guide/guide.md
burdges: These parachain validator groups are very small so repeating the `CandidateReceipt`s works fine.  It's their collection into one backed extrinsic that deduplicates this data.

docs/polkadot/implementors-guide/guide.md
rphmeier: It doesn't seem like we need a `CandidateAvailable` extrinsic since availability is computed implicitly on processing bitfields. Furthermore we want to process all bitfields at once since interleaving sounds complicated. I'm still in favor of bundling them since it doesn't seem we lose much by it

docs/polkadot/implementors-guide/guide.md
rphmeier: No, stuff like voting `Valid`/`Invalid` on the same block or seconding two candidates under the same relay-parent.

docs/polkadot/implementors-guide/guide.md
rphmeier: It's currently described in the overseer section below

docs/polkadot/implementors-guide/guide.md
```diff
@@ -0,0 +1,1013 @@
...
+
+during the block:
+  * Receive availability bitfields and move candidates from a pending availability to included state. See subsection below
+  * Receive new backed candidates to target for availability. See subsection below.
+  * Receive updates to configuration.
+
+process availability bitfields:
+  * We will accept an optional signed bitfield from each validator in each block.
+  * We need to check the signature and length of the bitfield for validity. 
+  * We will keep the most recent bitfield for each validator in the session. Each bit corresponds to a particular parachain candidate pending availability. Parachains are scheduled on every block, so we can assign a bit to each one of those. Parathreads are not scheduled on every block, and there may be a lot of them, so we probably don't want a dedicated bit in the bitfield for those. Since we want an upper bound on the number of parathreads we have scheduled or pending availability, a concept of "execution cores" used in scheduling (TODO) should be reusable here - have a dedicated bit in the bitfield for each core, and each core will be assigned to a different parathread over time.
```
burdges, line 306: We're currently discussing on riot if "core" should actually refer to a bit in the bitfield:  Rob suggested that parachains should've their own bit so that they cannot be blocked on a bit being freed up.  In that case, parachains get scheduled onto bits for the entire session, but parachain validator groups rotate with which bits they vote.  We could've physical cores consisting of parachain validator groups and logical cores consisting of bits in the bit field.

burdges: We're currently thinking parathreads and parachains behave opposite here, just so that only one thing is rotating in each case.

docs/polkadot/implementors-guide/guide.md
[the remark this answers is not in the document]
burdges: I've filed fishermen under the provider bitfield game described in https://github.com/paritytech/polkadot/pull/2089#issuecomment-740503318 and now https://github.com/w3f/research-security-issues/issues/30#issuecomment-768593265

PR #85, opened 2020-05-11
A&V networking: sentry node proxies

infinity0:
As discussed previously, support direct sends for sentry nodes. The extension is pretty straightforward, and replaces the previous more-complex extension on arbitrary network reachabilities.

### DIFF
--- a/docs/polkadot/networking/3-avail-valid.md
+++ b/docs/polkadot/networking/3-avail-valid.md
@@ -86,15 +86,15 @@ Recall that we have a disjoint partition of N validators into C sets of parachai
 
 The random assignment must be a deterministic but unpredictable assignment that every validator can calculate in the same way. For this purpose, we can use some entropy extracted from the chain at a position (height) determined by the current period, to seed a deterministic shuffle algorithm across the entire set of validators. The actual input seed must be pre-processed from the on-chain entropy, e.g. via HKDF, such that it is not re-used in any other security context.
 
-Example:
+**Example**: Let's say we have 20 validators `[a, b, c, ..., t]` and 5 parachains. The co-ordinates of the validators could look like such `[a: (0, 0), b: (0, 1), c: (0, 2), d: (0, 3), e: (1, 0), ..., t: (4, 3)]`. The validator set of the first parachain would be `[a, b, c, d]`. The first validator ring would be `[a, e, i, m, q]`, the second `[b, f, j, n, r]`.
 
-Let's say we have 20 validators `[a, b, c, ..., t]` and 5 parachains. The co-ordinates of the validators could look like such `[a: (0, 0), b: (0, 1), c: (0, 2), d: (0, 3), e: (1, 0), ..., t: (4, 3)]`. The validator set of the first parachain would be `[a, b, c, d]`. The first validator ring would be `[a, e, i, m, q]`, the second `[b, f, j, n, r]`.
+We assume that all nodes can reach each other directly via the physical topology. (For an extension that relaxes this assumption, see [Sentry node proxies](#sentry-node-proxies)). In terms of actual communication, a validator will be expected to communicate with:
 
-A validator ring is mostly-connected as permitted by the physical topology. Nodes within this ring talk to each other periodically via short-term and low-cost QUIC connections.
+- all other validators in their "ring"
+- all other validators in the same preliminary-check set; this is the same as the parachain validator set
+- all other validators in the same approval-checking set
 
-For a given parachain, the preliminary checkers are also mostly-connected, as permitted by the physical topology. Likewise the approval checkers are also mostly-connected.
-
-These connections represent the vast majority of traffic flow in our A&V networking protocol; to improve reliability and availability there are also other lines of communication as described below.
+This communication is to be done via short-term [QUIC](https://quicwg.org/base-drafts/draft-ietf-quic-transport.html) connections. These have a low connection setup latency (0- or 1-RTT), and maintaining a connection also uses up no OS-level resources. So it is generally unproblematic to have a few hundred of them. This communication represents the majority of traffic flow in our A&V networking protocol.
 
 The protocol runs in several phases and stages. Every node acts both in the distributor and distributee role, but not every role is active in every stage. A summary follows:
 
--- a/docs/polkadot/networking/3-avail-valid.md
+++ b/docs/polkadot/networking/3-avail-valid.md
@@ -271,4 +271,4 @@ The proxying protocol is straightforward, since the private validator node and t
 
 Recall that as above, there are two types of outbound connections: distributors pushing data, or distributees requesting data.
 
-Since the private validator node may not be able to access the address book, the sentry node is the one to perform the address book lookup. As described in 1(a) above, in the general case it will get a set of addresses as the result. For better load-balancing, the sentry node should sort this set and select the jth address to connect to, where j = i mod n, n is the size of the set, and (c, i) is the co-ordinate of its validator.
+Since the private validator node may not be able to access the address book (e.g. one implemented via Kademlia DHT), the sentry node is the one to perform the address book lookup. As described in 1(a) above, in the general case it will get a set of addresses as the result. For better load-balancing, the sentry node should sort this set and select the jth address to connect to, where j = i mod n, n is the size of the set, and (c, i) is the co-ordinate of its validator.
--- a/docs/polkadot/networking/3-avail-valid.md
+++ b/docs/polkadot/networking/3-avail-valid.md
@@ -19,6 +19,8 @@ Note: data from the relay chain is fully-replicated at each node, outside of thi
 ## TODO
 
 - to save time, the initial implemented version will be via gossip. Make a note of this.
+- handle the case where C does not evenly divide N
+- mention alternate topology suggested by jeff, more general & connected (in some sense) but also more complex
 - rough performance analysis & implementation notes, at the end of the doc.
 
 ## Background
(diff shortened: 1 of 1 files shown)


infinity0:
cc @mxinden

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -230,57 +233,42 @@ When receiving each piece, we also need to be able to authenticate it individual
...
+It's unnecessary to distinguish between "is a validator" and "acting as a proxy for a validator". This could be given as optional information in the certificate (e.g. so that the peer expects a higher latency), or it may be omitted if the validator wants to withhold this information from its peers. Nodes are free to guess whether their peers are proxies or not.
 
-If a proxy is chosen successfully, this should be remembered for the next period and the proxy may be re-used, skipping any failed attempts to send directly or via other proxies. Of course, if this proxy subsequently fails then we can remove this association and retry the steps from scratch.
+#### Proxy protocol
 
-The above applies for all stages where distributors are active. As per the [summary table](#Protocol-overview) this is all stages, except phase 2 stage B.
+The proxying protocol is straightforward, since the private validator node and the sentry nodes trust each other.
 
-**Distributees**
+1. Inbound, the protocol does not require any special headers (unwrapping/rewrapping of the content). Whenever a sentry node accepts an incoming connection, it forwards it directly onto the single sentry node.
```
mxinden, line 266:
```suggestion
1. Inbound, the protocol does not require any special headers (unwrapping/rewrapping of the content). Whenever a sentry node accepts an incoming connection, it forwards it directly onto the single validator node.
```

Correct?

mxinden, line 266:
```suggestion
1. Inbound, the protocol does not require any special headers (unwrapping/rewrapping of the content). Whenever a sentry node accepts an incoming connection, it forwards it directly onto the corresponding validator node.
```

Today a sentry node can sentry for multiple validator nodes. I guess, given that the original message includes the validator session key (or some similar identification token), the sentry node can decide which validator to forward the message to.

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -230,57 +233,42 @@ When receiving each piece, we also need to be able to authenticate it individual
...
+2. Outbound, the protocol needs special headers for the private validator node to tell the sentry node the outgoing destination. This is straightforward: namely the peer validator's key.
 
-For phase 2 stage B, we have N possible options to ask for ceil(N/3) pieces.
+Recall that as above, there are two types of outbound connections: distributors pushing data, or distributees requesting data.
 
-These numbers are high enough that we consider it unnecessary to specify that requests-for-pieces can be proxied, which itself is also slightly more complex than proxying a push as it involves one extra half-round of messages. This allows us to avoid some (hopefully) unnecessary complexity in what is already a fairly complex protocol.
+Since the private validator node may not be able to access the address book, the sentry node is the one to perform the address book lookup. As described in 1(a) above, in the general case it will get a set of addresses as the result. For better load-balancing, the sentry node should sort this set and select the jth address to connect to, where j = i mod n, n is the size of the set, and (c, i) is the co-ordinate of its validator.
```
mxinden, line 274: > jth address

What does *jth* stand for?

mxinden, line 274:
```suggestion
Say the address book is realized via a Kademlia Dht. Given that the private validator won't be able to make outbound connections beyond its sentry nodes the validator node may not be able to access the address book. The sentry node is the one to perform the address book lookup. As described in 1(a) above, in the general case it will get a set of addresses as the result. For better load-balancing, the sentry node should sort this set and select the jth address to connect to, where j = i mod n, n is the size of the set, and (c, i) is the co-ordinate of its validator.
```

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -230,57 +233,42 @@ When receiving each piece, we also need to be able to authenticate it individual
...
+1. For the address book, the validator should insert (or have their sentry nodes insert) the addresses of whatever nodes are acting as incoming proxies for it, what other people can reach. The following details are important:
 
-   k choices are to be tried from each set, before moving onto the next set; k = 3 seems like a reasonable *a priori* choice without real-world data.
+    a. each record should include a creation date, so that later entries unambiguously obsolete earlier entries. Thus load-balancing can be done in a more predictable way, across the full set of addresses.
 
-   The choice should prioritise proxies with known shared reachability to t.
+    b. each record should include an expiry date, so that old addresses are unambiguously avoided by readers unsure if the entry they got is "too old" or not
 
-A proxy has "known shared reachability" to a target t iff:
+2. For transport session keys, any node claiming to be a validator or a proxy for one, must present a certificate proving that the validator key authorises the transport key to do so. This ties back into the [authentication proposals](#L-authentication.html#proposal-fresh-authentication-signals).
```
mxinden, line 258: These *certificates* would tie into the proposal in https://github.com/w3f/research/blob/master/docs/polkadot/networking/L-authentication.rst correct?

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -8,14 +8,18 @@
 
 # Availability and Validity (A&V) networking
 
-Status: draft; accepted with a few TODO items remaining
+This subprotocol occurs whenever the block production protocol has output a candidate block for the current relay chain height.
```
mxinden, line 11:
```suggestion
This subprotocol occurs whenever the relay chain block production protocol has output a candidate block for the current relay chain height.
```

docs/polkadot/networking/3-avail-valid.md
infinity0: It defines `j` just a few words afterwards, is it not clear?

docs/polkadot/networking/3-avail-valid.md
infinity0: Yes, there is a link right there :)

docs/polkadot/networking/3-avail-valid.md
infinity0: I wanted this paragraph to be general, since in the future we may want to have part of the address-book onchain in which case the private validator node would be able to access it. But still part of it would be on the kademlia DHT, so it is probably better to have the sentry node resolve it all the time.

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -63,23 +67,24 @@ A4: TODO: Protocols should not be easily spammable, or the spammers should be ea
...
 2. receipts of "enough" pieces ("I have >1/3 of pieces, I don't need more"). This could also double as an attestation to everyone else, that there is availability from the attester's position on the network.
-3. [TBX S1 #1]
 
-This metadata should be gossiped every few seconds. The data of the actual pieces are distributed via a separate topology as described below:
+These should be gossiped every few seconds, and allows the participants to know when the stages of the protocol begin and end, details below.
+
+The data of the actual pieces are distributed via a topology separate from the broadcast medium:
 
 Recall that we have a disjoint partition of N validators into C sets of parachain validators, each set having size N/C. For our purposes for this subprotocol, we will randomly assign a co-ordinate (c, i) to every validator, with c in [0, C) and i in [0, N/C). Fixing c and varying i defines a particular parachain validator set; varying c and fixing i defines what we'll call a particular validator "ring". This name is only meant to be very slightly suggestive, the precise structure and its justification will be described below.
```
rphmeier, line 85: This is beyond the scope of the current PR, but I wonder about the coordinate scheme in situations where we may have groups of slightly different sizes if `C` is not a divisor of `N`. In practice, we have two options. The first is to spread the remainder over groups, so you have `N%C` groups with 1 extra validator. The other option is to leave some random N%C validators idle for a session. Ensuring that `C` is a divisor poses challenges on implementation and usability so I'd prefer to leave that off the table. Our thinking has been leaning towards option 1 so far, what do you think?

burdges: Yes, one extra sounds fine.  I still kinda dislike the grid topology here since we can attach fairly regular bipartite graphs every parachain group pair.

infinity0: The grid topology was mainly motivated back when originally I thought I might have to handle nodes with arbitrary NAT situations. So it is not essential, it can be changed. However it does make load-balancing easy to reason about, and I couldn't immediately think of anything that was significantly better.

infinity0: I will add a paragraph mentioning what happens with unevenly-sized parachains, with the "one extra" approach. BTW, this problem occurs with the bipartite-graphs topology as well, since obviously you can't have a perfect matching between sets of size (C) and (C+1).

docs/polkadot/networking/3-avail-valid.md
infinity0: Added, thanks

docs/polkadot/networking/3-avail-valid.md
infinity0: OK, this is actually way less trivial than I thought, at leas to do it in a way that preserves load-balancing properties and avoids some nodes having to do 2x the amount of work. Will have to spend a bit more time to think about this.

infinity0: Note: simply spreading the extra validators across the existing groups, is not well-defined, because you need to define what pieces they are supposed to distribute to everyone else in the group. But there is already someone from their chain, that is distributing a well-defined set of pieces to everyone else in the group.

I will merge this PR and file an issue about this.

infinity0: Tracking in w3f/research-internal#390

PR #86, opened 2020-05-18
clean up module boundaries and optimize GC

rphmeier:
1. Introduces an `InclusionInherent` module specifically for negotiating between the `Scheduler` and the `Inclusion` modules for availability bitfields, re-scheduling, timing out occupied cores, and then including backed candidates. This PR almost breaks the coupling between the modules entirely and the approach could be used to entirely uncouple `Inclusion` from `Scheduler` and perhaps `Paras` as well.
1. Uses this division to time out occupied cores directly after availability, not in `Scheduler` finalization. This is slightly more efficient at minimizing lost time by honest validator groups, as it saves a block duration on timeouts.

### DIFF
--- a/docs/polkadot/implementors-guide/guide.md
+++ b/docs/polkadot/implementors-guide/guide.md
@@ -709,7 +709,7 @@ Actions:
 * `occupied(Vec<CoreIndex>). Note that the given cores have become occupied. This clears them from `Scheduled`. Fails if any given cores were not scheduled.
 * `core_para(CoreIndex) -> ParaId`: return the currently-scheduled or occupied ParaId for the given core.
 * `group_validators(GroupIndex) -> Vec<ValidatorIndex>`
-* `availability_timeout_predicate() -> Option<impl Fn(CoreIndex, BlockNumber) -> bool>`: returns an optional predicate that should be used for timing out occupied cores. if `None`, no timing-out should be done. The predicate accepts the index of the core, and the block number since which it has been occupied. The predicate should be implemented based on the time since the last validator group rotation, and the respective parachain and parathread timeouts, i.e. only within `max(config.chain_availability_period, config.thread_availability_period)`
+* `availability_timeout_predicate() -> Option<impl Fn(CoreIndex, BlockNumber) -> bool>`: returns an optional predicate that should be used for timing out occupied cores. if `None`, no timing-out should be done. The predicate accepts the index of the core, and the block number since which it has been occupied. The predicate should be implemented based on the time since the last validator group rotation, and the respective parachain and parathread timeouts, i.e. only within `max(config.chain_availability_period, config.thread_availability_period)` of the last rotation would this return `Some`.
 
 ### The Inclusion Module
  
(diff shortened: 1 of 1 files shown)


docs/polkadot/implementors-guide/guide.md
```diff
@@ -619,7 +619,7 @@ Validator group assignments do not need to change very quickly. The security ben
...
+When a rotation occurs, validator groups are still responsible for distributing availability pieces for any previous cores that are still occupied and pending availability. In practice, rotation and availability-timeout frequencies should be set so this will only be the core they have just been rotated from. It is possible that a validator group is rotated onto a core which is currently occupied. In this case, the validator group will have nothing to do until the previously-assigned group finishes their availability work and frees the core or the availability process times out. Depending on if the core is for a parachain or parathread, a different timeout `t` from the `HostConfiguration` will apply. Availability timeouts should only be triggered in the first `t-1` blocks after the beginning of a rotation.
```
burdges, line 622: Is there a missing "in practice" the last sentence?  We could've a validator group with enough crappy validators who fail (a) distributing pieces, or (b) providing the block to stronger validator group members, but a subsequent luckier collator have more luck reaching the stronger validator group members.

rphmeier: This was Al's suggestion. If you have enough honest validators in your group to back a candidate, you shouldn't have any trouble distributing pieces, right?

burdges: I'll ask him.  I think dishonest or crappy validators could halt the parachain for one rotation this way.

AlistairStewart: If the group has enough crappy validators, then it can always halt block production for a rotation. The goal here was to try to ensure that they couldn't tie it up for any more than a rotation by getting the new validators to validate as fast as possible.

We need at least of majority of a group to back a candidate; I don't know if we ever decided on the exact number here. But if any one of these is competetent, they should be able to do (b), whereas (a) might need several competent validators. 

So we should be able to assume that the group collectively is crappy and try to get the next guys on the job as fast as possible.

AlistairStewart: We should focus our efforts on dealing with this kind of situation if it turns out to happen frequently by fixing networking on the validator and collator side. Timing out availability might just give us the same problem on the next block, since it is the same group of validators and the backing validators will overlap a lot.

burdges: Is there any code simplification or other benefit in constraining when availability timeouts occur though?  

At first I'd imagined some high proportion like 2/3 solves my concern, but If all but one validator turn out crappy then allowing later timeouts improves performance. 

We've mostly discussed 1/2 but actually 2/3 of the group ensures that 2/3 of validators can contact the group, assuming no NATs, etc.  I think high like 2/3 ensure availability runs faster, while lower makes availability start slower, but actually gives availability more running time.

rphmeier: It's slightly cheaper to avoid doing a GC sweep every block. We're currently doing this in an O(N) way, but we could also trade off more storage and build an index for knowing better which cores need to be collected.

I lean towards the fewer timeouts approach for that reason and the fact that there are other ways to halt parachain inclusion for a rotation. Rotations are relatively fast, so only timing out recently after rotations wouldn't be a huge issue.

I've been assuming a simple majority, not a supermajority for candidate groups, because liveness is probably a bigger issue than safety in the early versions of Polkadot. We can tweak this later if we want.

PR #87, opened 2020-05-19
Expand Candidate Selection section

coriolinus:
Note: there are a bunch of whitespace edits where my editor stripped trailing whitespace; I can edit to leave those in, but IMO it's cleaner to just fix it.

### DIFF
--- a/docs/polkadot/implementors-guide/guide.md
+++ b/docs/polkadot/implementors-guide/guide.md
@@ -913,7 +913,7 @@ Furthermore, the protocols by which subsystems communicate with each other shoul
 
 #### Description
 
-The Candidate Backing subsystem determines whether parachain blocks are valid or not. When run as a validator or fisherman, this subsystem actually validates incoming candidates.
+The Candidate Backing subsystem ensures at least one preliminary validator commits to each parablock's correctness. Parablocks for which no validator will assert correctness are discarded. If the block later proves invalid, the initial backers are slashable; this gives polkadot a rational threat model during subsequent stages.
 
 Its role is to produce backed candidates for inclusion in new relay-chain blocks. It does so by issuing signed [Statements](#Statement-type) and tracking received statements signed by other validators. Once enough statements are received, they can be combined into backing for specific candidates.
 
--- a/docs/polkadot/implementors-guide/guide.md
+++ b/docs/polkadot/implementors-guide/guide.md
@@ -947,7 +947,7 @@ This will probably blur the lines between jobs, will probably require inter-job 
 
 The Candidate Backing Job represents the work a node does for backing candidates with respect to a particular relay-parent.
 
-The goal of a Candidate Backing Job is to produce as many backed candidates as possible. This is done via signed [Statements](#Statement-type) by validators. If a candidate receives a majority of supporting Statements from the Parachain Validators currently assigned, then that candidate is considered backed.
+The goal of a Candidate Backing Job is to produce as many accurate [Statements](#Statement-type) about candidates as possible. If a candidate receives a majority of supporting Statements from the Parachain Validators currently assigned, then that candidate is considered backed.
 
 *on startup*
 * Fetch current validator set, validator -> parachain assignments from runtime API.
--- a/docs/polkadot/implementors-guide/guide.md
+++ b/docs/polkadot/implementors-guide/guide.md
@@ -917,8 +917,6 @@ The Candidate Backing subsystem determines whether parachain blocks are valid or
 
 Its role is to produce backed candidates for inclusion in new relay-chain blocks. It does so by issuing signed [Statements](#Statement-type) and tracking received statements signed by other validators. Once enough statements are received, they can be combined into backing for specific candidates.
 
-It also detects double-vote misbehavior by validators as it imports votes, passing on the misbehavior to the correct reporter and handler.
-
 #### Protocol
 
 The **Candidate Selection** subsystem is the primary source of non-overseer messages into this subsystem. That subsystem generates appropriate [`CandidateBackingSubsystemMessage`s](#Candidate-Backing-Subsystem-Message), and passes them to this subsystem.
@@ -1008,8 +1006,10 @@ This module is only ever interested in parablocks assigned to the particular par
 New parablock candidates may arrive from a potentially unbounded set of collators. This subsystem chooses either 0 or 1 of them per relay parent to second. If it chooses to second a candidate, it sends an apropriate message to the **Candidate Backing** subsystem to generate an appropriate `Statement`.
 
 All parablocks which peers have seconded are also sent to the Candidate Backing subsystem for re-validation.
+As seconded peers are tallied, double-votes are detected. If found, a report is sent to the **Misbehavior Arbitration** subsystem.
 
 In the event that a parablock candidate proves invalid, this subsystem will receive a message back from the Candidate Backing subsystem indicating so. If that parablock candidate originated from a collator, this subsystem will blacklist that collator. If that parablock candidate originated from a peer, this subsystem generates a report for the **Misbehavior Arbitration** subsystem.
+
 
 TODO: more details, protocol, etc
 
--- a/docs/polkadot/implementors-guide/guide.md
+++ b/docs/polkadot/implementors-guide/guide.md
@@ -138,14 +138,14 @@ In this example, group 1 has received block C while the others have not due to n
    .        (Validator 5)             .
    .                                  .
    ....... Building on C and C' .......
-           
-            +----------------------+         +----------------------+       
-            |    Relay Block C     |         |    Relay Block C'    |          
+
+            +----------------------+         +----------------------+
+            |    Relay Block C     |         |    Relay Block C'    |
             +----------------------+         +----------------------+
                             \                 /
-                             \               /   
+                             \               /
                               \             /
-                      +----------------------+    
+                      +----------------------+
                       |  Relay Block B       |
                       +----------------------+
 	                             |
@@ -181,7 +181,7 @@ Our Parachain Host includes a blockchain known as the relay-chain. A blockchain 
                                 +----------------+
                                 |    Genesis     |
                                 +----------------+
-```                                
+```
 
 
 A blockchain network is comprised of nodes. These nodes each have a view of many different forks of a blockchain and must decide which forks to follow and what actions to take based on the forks of the chain that they are aware of.
@@ -235,7 +235,7 @@ It is also helpful to divide Node-side behavior into two further categories: Net
                                                    ______| |______
                                                    ___Transport___
 
-```                                                   
+```
 
 
 Node-side behavior is split up into various subsystems. Subsystems are long-lived workers that perform a particular category of work. Subsystems can communicate with each other, and do so via an Overseer that prevents race conditions.
@@ -292,7 +292,7 @@ initialization:
   * determine scheduled parachains and parathreads for the upcoming block or blocks.
   * determine validator assignments to scheduled paras for the upcoming block or blocks.
   * remove blocks which have been pending availability for too long. this is tightly coupled with scheduling.
-  * handle the start of a new session - discard all candidates pending availability and note the upcoming validator set. 
+  * handle the start of a new session - discard all candidates pending availability and note the upcoming validator set.
   * apply calls from upward messages - messages from parachains to the relay chain.
 
 during the block:
@@ -302,7 +302,7 @@ during the block:
 
 process availability bitfields:
   * We will accept an optional signed bitfield from each validator in each block.
-  * We need to check the signature and length of the bitfield for validity. 
+  * We need to check the signature and length of the bitfield for validity.
   * We will keep the most recent bitfield for each validator in the session. Each bit corresponds to a particular parachain candidate pending availability. Parachains are scheduled on every block, so we can assign a bit to each one of those. Parathreads are not scheduled on every block, and there may be a lot of them, so we probably don't want a dedicated bit in the bitfield for those. Since we want an upper bound on the number of parathreads we have scheduled or pending availability, a concept of "execution cores" used in scheduling (TODO) should be reusable here - have a dedicated bit in the bitfield for each core, and each core will be assigned to a different parathread over time.
   * Bits that are set to `true` denote candidate pending availability which are believed by this validator to be available.
   * Candidates that are pending availability and have the corresponding bit set in 2/3 of validators' bitfields (only counting those submitted after the candidate was included, since some validators may not have submitted bitfields in some time) are considered available and are then moved into the "pending approval" state.
@@ -316,8 +316,8 @@ candidates entering the "pending approval" state:
 
 process new backed candidates:
   * ensure that only one candidate is backed for each parachain or parathread
-  * ensure that the parachain or parathread of the candidate was scheduled and does not currently have a block pending availability. 
-  * check the backing of the candidate. 
+  * ensure that the parachain or parathread of the candidate was scheduled and does not currently have a block pending availability.
+  * check the backing of the candidate.
   * move to "pending approval" state. (pass along any configuration information that is liable to change)
 
 misbehavior reports and secondary checks:
@@ -366,7 +366,7 @@ The Initializer module is special - it's responsible for handling the initializa
 The Parachain Host operates under a changing set of validators. Time is split up into periodic sessions, where each session brings a potentially new set of validators. Sessions are buffered by one, meaning that the validators of the upcoming session are fixed and always known. Parachain Host runtime modules need to react to changes in the validator set, as it will affect the runtime logic for processing candidate backing, availability bitfields, and misbehavior reports. The Parachain Host modules can't determine ahead-of-time exactly when session change notifications are going to happen within the block (note: this depends on module initialization order again - better to put session before parachains modules). Ideally, session changes are always handled before initialization. It is clearly a problem if we compute validator assignments to parachains during initialization and then the set of validators changes. In the best case, we can recognize that re-initialization needs to be done. In the worst case, bugs would occur.
 
 There are 3 main ways that we can handle this issue:
-  1. Establish an invariant that session change notifications always happen after initialization. This means that when we receive a session change notification before initialization, we call the initialization routines before handling the session change. 
+  1. Establish an invariant that session change notifications always happen after initialization. This means that when we receive a session change notification before initialization, we call the initialization routines before handling the session change.
   2. Require that session change notifications always occur before initialization. Brick the chain if session change notifications ever happen after initialization.
   3. Handle both the before and after cases.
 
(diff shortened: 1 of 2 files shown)


docs/polkadot/implementors-guide/guide.md
```diff
@@ -994,9 +992,44 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
 (TODO: send statements to Statement Distribution subsystem, handle shutdown signal from candidate backing subsystem)
 
+### Candidate Selection Subsystem
+
+#### Description
+
+The Candidate Selection subsystem monitors net traffic for two events:
+
+- a new parablock candidate is available
+- a peer has seconded a parablock candidate
```
rphmeier, line 1002: The reason for this is that we'd prefer to avoid seconding the same parablock as another peer?

Given our short timeframe here, it might not be the case that higher candidate diversity is actually better.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -994,9 +992,44 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+
+This module is only ever interested in parablocks assigned to the particular parachain which this validator is currently handling.
+
+New parablock candidates may arrive from a potentially unbounded set of collators. This subsystem chooses either 0 or 1 of them per relay parent to second. If it chooses to second a candidate, it sends an apropriate message to the **Candidate Backing** subsystem to generate an appropriate `Statement`.
+
+All parablocks which peers have seconded are also sent to the Candidate Backing subsystem for re-validation.
+As seconded peers are tallied, double-votes are detected. If found, a report is sent to the **Misbehavior Arbitration** subsystem.
```
rphmeier, line 1009: There is a division to be clarified about which subsystems are responsible for double-vote reporting. Candidate Backing, Statement Distribution, and Candidate Seconding subsystems are all going to be in position to detect double-vote misbehavior. We shouldn't be filing duplicate reports from all systems, since one should be sufficient. Of these, I think Statement Distribution is probably the best fit, as it's also the system that's capable of dropping new incoming statements - the marginal cost of double-voting more than once is 0, so someone who double-votes is likely to spam.

We could establish an invariant that statement distribution doesn't give other subsystems double-votes.

docs/polkadot/implementors-guide/guide.md
coriolinus: Yes, I'd had the impression that it was actually a slashable error to second the same parablock as a peer; the best we can do in that case is mark it as `Valid`.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -994,9 +992,44 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+The Candidate Selection subsystem monitors net traffic for two events:
+
+- a new parablock candidate is available
+- a peer has seconded a parablock candidate
+
+This module is only ever interested in parablocks assigned to the particular parachain which this validator is currently handling.
+
+New parablock candidates may arrive from a potentially unbounded set of collators. This subsystem chooses either 0 or 1 of them per relay parent to second. If it chooses to second a candidate, it sends an apropriate message to the **Candidate Backing** subsystem to generate an appropriate `Statement`.
+
+All parablocks which peers have seconded are also sent to the Candidate Backing subsystem for re-validation.
```
rphmeier, line 1008: The Candidate Backing subsystem should be getting `Seconded` statements from the statement distribution subsystem (not yet defined), so sending them again here seems redundant. 

This makes me think we want another subsystem that actually does validation, because there are going to be situations like these where different subsystems validate candidates and we don't want to do the work twice: this one, which validates candidates received from collators, and the candidate backing system which validated candidates received from the network.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -994,9 +992,44 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+
+### Misbehavior Arbitration Subsystem
+
+#### Description
+
+The Misbehavior Arbitration system collects reports of validator misbehavior, and slashes the stake of both misbehaving validator nodes and false accusers.
+
+#### TODOs
+
+- threshold of reports for deciding that a validator has misbehaved
```
rphmeier, line 1024: A single double-vote report is full evidence, so that should be enough.

docs/polkadot/implementors-guide/guide.md
coriolinus: OK, I don't feel strongly about which of those subsystems has the formal responsibility to detect double-votes. Agree that only one of them should really be in charge of it.

Not sure how to establish that invariant: couldn't we get Vote 1, distribute it, then receive Vote 2 from the same source? We can report double-voting at that point, and we can cease to distribute Vote 1, but we can't rescind those instances we've already sent.

docs/polkadot/implementors-guide/guide.md
coriolinus: Heh: I'd thought of breaking out a Candidate Validation subsystem, but I couldn't convince myself that it was sufficiently distinct from Candidate Backing to be worth it. That said, I'm perfectly willing to accept that distinction.

I think that we-the-team should probably have a verbal discussion in the near future to decide the rough overview of the enumeration of subsystems, and the major responsibilities of each. Otherwise we'll keep running into minor divergence of assumptions like this.

docs/polkadot/implementors-guide/guide.md
coriolinus: What's the distinction between Candidate Backing and Candidate Seconding, by the way? The latter is new to me.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -913,17 +913,15 @@ Furthermore, the protocols by which subsystems communicate with each other shoul
 
 #### Description
 
-The Candidate Backing subsystem is engaged in by validators in to contribute to the backing of parachain candidates submitted by other validators.
+The Candidate Backing subsystem determines whether parachain blocks are valid or not. When run as a validator or fisherman, this subsystem actually validates incoming candidates.
```
burdges, line 916: It's true they validate the block, but it's missleading this way.  Instead I'd suggest:

"The Candidate Backing subsystem ensures at least one preliminary validator commits to each parablock's correctness.  These initial backers being slashable if the block later proves invalid gives polkadot a rational threat model during subsequent stages."

docs/polkadot/implementors-guide/guide.md
burdges: We do not afaik require double vote defenses:  

We should prevent a parablocks' first backer from being first backer for another parablock until it either progresses to availability or expires, but that's only so that one bad first backer cannot halt a parachain for one rotation, and the logic appears easy since replacing an expired candidate receipt reveals the previous first backer.  

We could encourage duplicate backing votes on competing parablock because the relay chain block producer chooses among them definitively.  Availability and approval votes cannot be duplicates by construction.

docs/polkadot/implementors-guide/guide.md
burdges: Yes, anytime any two validators disagree about a parablock, then we immediately escalate to all validators checking, and the minority side is slashed 100%.

docs/polkadot/implementors-guide/guide.md
burdges: In particular, our first validator produces lite client proofs about the relay chain parent, which later validators only check, so that requires some difference in execution context I guess.

docs/polkadot/implementors-guide/guide.md
burdges: We do not afaik require slashing here.  see https://github.com/w3f/research/pull/87/files#r427365330

docs/polkadot/implementors-guide/guide.md
rphmeier: While true, I've assumed we only need to do those escalations for blocks that going through the approval process. Full escalation during the backing phase is probably unnecessary.

docs/polkadot/implementors-guide/guide.md
rphmeier: > In particular, our first validator produces lite client proofs about the relay chain parent, which later validators only check, so that requires some difference in execution context I guess.

Well, maybe in XCMP. Not with the state of the doc as-is

docs/polkadot/implementors-guide/guide.md
rphmeier: What does it mean to be the "first backer"? Ordering is really relative to network conditions. This first-backer constraint sounds complicated and don't seem to get us much, as the para will still be live upon the next rotation. Double-votes are the only misbehavior we want to want to discourage at this phase as everything else is covered by the approval system, which hasn't been specified yet. It's easy to avoid double-voting as an honest validator and the slashing condition is strict.

docs/polkadot/implementors-guide/subsystems-overview.md
```diff
@@ -0,0 +1,95 @@
...
+### Outbound Messages
+
+- **Candidate Validation**: double-check this peer's seconded candidate
+- **Peer validators**: Here's what I think about a candidate's validity
+- **Misbehavior Arbitration**: I disagree with a peer node about this candidate's validity
+- **Overseer**: a unanimous quorum of nodes has agreed about this candidate's validity
+
+## Misbehavior Arbitration
+
+The misbehavior arbitration module kicks in when two validator nodes disagree about a candidate's validity. In this case, _all_ validators, not just those assigned to its parachain, weigh in on the validity of this candidate. The minority is slashed.
```
rphmeier, line 73: Why not have a specific misbehavior system for double-votes? The dispute between valid/invalid would be handled mostly by the `Validity` module, a runtime API, and a `SecondaryChecking` subsystem.
1. Validity module tracks blocks currently up for approval, backing, output of all secondary checks so far. it handles slashing when there are disputes. 
2. Runtime API makes all information available to node-side
3. SecondaryChecks uses information about blocks up for approval (random seed, escalation) to schedule and perform secondary checks.

docs/polkadot/implementors-guide/guide.md
rphmeier: BTW, by double-voting I've meant one of two things:
  - Issuing two `Seconded` statements.
  - Issuing conflicting statements on the same candidate (Valid/Invalid, Seconded/Valid, Seconded/Invalid)

docs/polkadot/implementors-guide/guide.md
burdges: I think escalation here improves security since an adversary could miss calculate their ability to limit the other parachain validators access.  We could postpone doing this escalation until later however since we do not consider this the primary mechanism.  @AlistairStewart ?  

I suppose excluding slashing from an entire module dramatically simplifies the code, yes?  After escalation works in approvals then we could consider abstracting that code to cover both, so maybe 2021..

docs/polkadot/implementors-guide/guide.md
burdges: Yes exactly.  I never even considered conflicting votes on the same candidate, but yes those are slashable.

docs/polkadot/implementors-guide/subsystems-overview.md
burdges: We always envisioned the validity module counting the parachain validator's backing votes btw, so yes maybe it could just handle their slashing along with its own.

coriolinus: This should not be a module, but a subsystem within the validity module; I misspoke. 

The point of Misbehavior Arbitration is that this role handles things when the whole voting set needs to vote on something: this is the subsystem which says "hey, peers: everyone judge this particular block", tallies the votes, and slashes the minority. That behavior is fairly complicated, so I thought it made more sense to put it into its own subsystem than to build it into a different subsystem with a different primary responsibility. 

In other words, assuming I understand the three-step checklist properly, I think Misbehavior Arbitration is the subsystem within the Validity module which handles step 1.

docs/polkadot/implementors-guide/guide.md
coriolinus: So, we agree that one validator `Second`ing two different parablocks with the same parents is a double-vote and should be slashed.

What I am still confused by: is it an error if two different validators both `Second` the same parablock? If yes: why, exactly; if there's a failure there, it's not obvious to me. If no: could we consider removing the `Seconded` variant from `Statement` and passing `CandidateReceipt`s via some other channel?

burdges: No.  A&V always extracts consensus from the relay chain to avoid engaging in its own BFT protocol within parachain validator groups. 

We want several parachain validators to second each candidate, perhaps 2/3rds, so if more than 1/3rd second two candidates then the parachain halts unless they can second more choices.

We can replace my suggestion to rotate "first backers" by merely asking that anytime we replace an expired candidate with signer set `S_old` by a new candidate with signer set `S_new` then `S_new` must include `min_new_backers` validators not in `S_old`.  We'd take `min_new_backers=0` if we asked for 2/3rds, but some parameter choices give `min_new_backers>0`.  We could then second as many parablock with the same parent as we like.

coriolinus: Thanks! In that case, I'd propose to simplify `Statement`s:

```rust
/// A statement about the validity of a parachain candidate
enum Statement {
    /// This parachain candidate is valid.
    Valid(CandidateReceipt),
    /// The parachain candidate with this hash is invalid.
    Invalid(Hash),
}
```

Motivation: eliminating the `Seconded` variant makes it much more clear what the semantics should be: a validator should never contradict its own statements about a block, and all validators should agree on the semantics. As far as I can tell, the only motivation for distinguishing the `Seconded` variant is to save some bytes for those validators who mark a block `Valid`. That feels like premature optimization, to me.

burdges: Yup, we forbid self conflicts.  :)  

I still think XCMP might distinguish between the first backer and later backers, meaning multiple parachain validators could compute different candidate receipts from the same parachain block, so "seconding" could refer to backing parachain validators after the first one.  I'd assumed you meant this initially, but we should avoid assuming a deterministic map from candidates to candidate receipts.

rphmeier: > As far as I can tell, the only motivation for distinguishing the Seconded variant is to save some bytes for those validators who mark a block Valid. That feels like premature optimization, to me.

Actually, it's to bound the number of candidates needed to validate, which is a legitimate DoS vector. The `Seconded` message gives us an upper bound on the number of candidates we'd ever need to validate or distribute during the backing phase, as long as nobody wants to get slashed 100%.

> We can replace my suggestion to rotate "first backers" by merely asking that anytime we replace an expired candidate with signer set S_old by a new candidate with signer set S_new then S_new must include min_new_backers validators not in S_old. We'd take min_new_backers=0 if we asked for 2/3rds, but some parameter choices give min_new_backers>0. We could then second as many parablock with the same parent as we like.

I'm still not sure why we want this - does it actually give us any more security, given that we will ensure the block is available and we have a secondary-checking system.

docs/polkadot/implementors-guide/guide.md
rphmeier: Yeah, I think this might be one for 2021. The first AnV push will ideally focus only on stuff that's made its way on-chain, and we can schedule another one later.

docs/polkadot/implementors-guide/subsystems-overview.md
```diff
@@ -0,0 +1,95 @@
...
+
+- **Overseer**: the majority of nodes agree that this candidate is valid/invalid; here is a list of minority voters to slash
+
+---
+
+## Tracing
+
+Let's follow a parachain candidate through the system. This follows the happy path, but we'll note exit points along the way.
+
+- Collator (external to this system)
```
rphmeier, line 90: I'd like collator subsystems to be covered here as well, since they're also block-based work.

docs/polkadot/implementors-guide/subsystems-overview.md
```diff
@@ -0,0 +1,95 @@
...
+- Overseer
+- Candidate Selection: every parachain validator can choose at most 1 candidate to back, so potentially many are discarded.
+- Candidate Backing
+- Statement Distribution: if there's disagreement about a candidate's validity, things move to misbehavior arbitration.
+- Overseer: a quorum of validators unanimously agrees about this candidate's validity.
```
rphmeier, line 95: why overseer here?

should probably go to a "block authorship" subsystem (although actually block authorship will happen outside of the overseer/subsystem hierarchy, I've been thinking we'll create a subsystem responsible for gathering information for that block authorship logic.

and from there it goes into the runtime.

docs/polkadot/implementors-guide/subsystems-overview.md
```diff
@@ -0,0 +1,95 @@
...
+- **Candidate Validation**: is this candidate in fact valid?
+- **Statement Distribution**: I testify that this candidate is (in-)valid
+
+## Candidate Validation
+
+The candidate validation subsystem handles the details of validating a candidate. For any given parachain candidate, it arranges the appropriate context: the validation function, the relay parent head data, the in-progress relay data, etc.
+
+### Inbound Messages
+
+- **Various Subsystems**: requests to validate a particular parablock candidate
```
rphmeier, line 48: one open question is how the `PoV` fetch is handled.

in the case of validating a block from a collator (to decide whether to second), the collator has sent us a PoV, so we could send it directly to the validation subsystem. In the case of validating a candidate seconded by another validator, we just have the receipt and not the PoV, so we'd want to proceed through these steps:
  1. Figure out if we've already validated the candidate
  2. If not, fetch PoV
  3. Upon PoV fetch, do validation

We might want something like a `PoVOrigin` sent alongside a validation request that tells us where to get the `PoV`. `Included` would be one option, and fetching from the network would be another. But maybe this is where things get complicated, since the network might want to use information about statements received from other validators to figure out who might have the PoV. At the moment, we'll just gossip all PoVs in the same subsystem as statements, so we can defer this to later.

docs/polkadot/implementors-guide/guide.md
rphmeier: @burdges and I talked separately, the issues are based on different mental models and are resolved by having collators fully responsible for generating PoVs, as they do in this document, so the first backer is not as important. We should keep the `Seconded` message and don't need to worry about the candidate-replacement signer overlap stuff as with this division between collation and validation it is a liveness concern as opposed to a safety one.

burdges: We figured this out.  We'll want a clearer deliniation between collation and backing than my notion of "first" gave.  It does not preclude later running the last phase of collation on validators, but this becomes an architectural question for cumulus, not A&V.  :)

burdges:
We've punted several things here: 
 - Escalation during backing until later - https://github.com/w3f/research/pull/87#discussion_r427373356
- Relay chain component of collation, including selecting relay parent and XCMP, to collation/cumulus, which also simplifies everything around what a PoV is - https://github.com/w3f/research/pull/87#discussion_r428731530

coriolinus:
Closing in favor of https://github.com/paritytech/polkadot/pull/1161 as the canonical location of the edited document has moved into that repo.

PR #88, opened 2020-05-20
A&V networking: change topology to one more suited to different chain-sizes

infinity0:
This new topology is closer to what @burdges was suggesting originally - a bit more complex, based on matching between pairs of chains, rather than the old ring-based one. However it's much easier to get working when C (num chains) does not evenly divide N (num validators).

There are a few similarities though, to help you understand it. When C does divide N, the ring-based one can appear as one special case of this new topology. Everyone has an in-neighbour set and an out-neighbour set, the difference is that this is not expected to be shared across chains. However, all the load-balancing logic can be re-written in terms of these neighbour sets, and that's also included in this PR.

I've defined the randomness seeds quite precisely, they might be different from jeff's original proposal but only because I can't remember what the details are. Variations on the general structure will all work, e.g. using some other symmetric function than XOR.

### DIFF
--- a/docs/polkadot/networking/3-avail-valid.md
+++ b/docs/polkadot/networking/3-avail-valid.md
@@ -33,7 +33,7 @@ Every block is erasure coded across N pieces with a threshold of ceil(N/3).
 
 At the start of the period, for every parachain, its N/C parachain validators each have all of the N pieces. In this role we call them the "preliminary checkers". The high-level purpose of A&V networking is to:
 
-R1: Distribute the pieces of all C blocks to all other validators. A corollary of this is that every validator *must receive* at minimum (C-1) pieces, 1 from every other parachain.
+R1: Distribute the pieces of all C blocks to all other validators. A corollary of this is that every validator *must receive* at minimum (C-1) pieces - 1 from every other parachain - when operating at full capacity i.e. when every C parachain produces a block.
 
 R2: Ensure that ceil(N/3) of the pieces of all C blocks remains available and retrievable for a reasonable amount of time, across the N validators.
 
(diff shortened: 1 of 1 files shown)


infinity0:
Some notes on why the old topology was hard to get working with uneven chain sizes, are here: research-internal/issues/390

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -35,9 +33,9 @@ Every block is erasure coded across N pieces with a threshold of ceil(N/3).
 
 At the start of the period, for every parachain, its N/C parachain validators each have all of the N pieces. In this role we call them the "preliminary checkers". The high-level purpose of A&V networking is to:
 
-1. Distribute the pieces of all C blocks to all other validators
+R1: Distribute the pieces of all C blocks to all other validators. A corollary of this is that every validator *must receive* at minimum (C-1) pieces, 1 from every other parachain.
```
rphmeier, line 36: nit: This is predicated on each parachain including a block, which is not guaranteed.

docs/polkadot/networking/3-avail-valid.md
```diff
@@ -75,28 +73,24 @@ To help implementation be divided into stages, the main proposal is defined with
...
-
-The random assignment must be a deterministic but unpredictable assignment that every validator can calculate in the same way. For this purpose, we can use some entropy extracted from the chain at a position (height) determined by the current period, to seed a deterministic shuffle algorithm across the entire set of validators. The actual input seed must be pre-processed from the on-chain entropy, e.g. via HKDF, such that it is not re-used in any other security context.
+The data of the pieces are distributed via the following communication links:
 
-**Example**: Let's say we have 20 validators `[a, b, c, ..., t]` and 5 parachains. The co-ordinates of the validators could look like such `[a: (0, 0), b: (0, 1), c: (0, 2), d: (0, 3), e: (1, 0), ..., t: (4, 3)]`. The validator set of the first parachain would be `[a, b, c, d]`. The first validator ring would be `[a, e, i, m, q]`, the second `[b, f, j, n, r]`.
+1. all validators in their in-neighbour set, as defined by in the overlay topology below
```
infinity0, line 87: Just realised this should say simply "neighbour sets" not just in-neighbour.

PR #89, opened 2020-05-21
Add a Statement Gossip Subsystem section to the implementors guide

expenses:


expenses:
@rphmeier @mxinden

docs/polkadot/implementors-guide/guide.md
```diff
@@ -989,6 +989,67 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+    overseer_sender: Sender<StatementGossipSubsystemOut>,
+    // A receiver of messages from the subsystem. This channel is exclusive to this job.
+    subsystem_receiver: Receiver<SignedStatement>,
+);
+```
+
+The `statement_gossip_job` should consist of a `futures::select!` block with 3 parts:
+
+* A future that takes the statements from the `subsystem_receiver` stream and gossips them along with the `relay_parent` hash using `gossip_service`.
+* A future that takes the messages that `gossip_service` receives for the `relay_parent`, filters to only the statements and semds them via `overseer_sender`.
```
rphmeier, line 1045:
```suggestion
* A future that takes the messages that `gossip_service` receives for the `relay_parent`, filters to only the statements and sends them via `overseer_sender`.
```

docs/polkadot/implementors-guide/guide.md
```diff
@@ -989,6 +989,67 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+The statement gossip subsystem is responsible for gossiping out and receiving [Signed Statements](#Signed-statement-type) from validators.
+
+This subsystem communicates with the overseer via [two message types](#Candidate-Backing-Subsystem-Messages), one for overseer -> subsystem and another for subsystem -> overseer. Currently this just differentiates between statements that the subsystem has been instructed to gossip out, and statements that the subsystem has received.
+
+Here is one design for how it might look in code:
+
+```rust
+struct StatementGossipSubsystem<GS: GossipService, SP: Spawn> {
+    /// This trait comes from the legacy networking code, so it is likely to be changed at some point.
+    gossip_service: GS,
```
rphmeier, line 1003: cc @mxinden I'd expect that we'd do things the way they're laid out here, where the subsystem owns its entire gossip service.

However with the gossip-service as-is we need to write a `GossipValidator` that handles incoming messages and checks signatures and controls what messages we send to peers. Since the validator set can change from block to block, we'll want to:
  - only accept messages related to the relay-parents of jobs that we have active
  - fetch the validator set for a parent-hash as part of the job-starting process - either by having overseer provide this in `StartWork` or by calling out to another `RuntimeApi` subsystem to avoid shared ownership of the runtime API.
  - communicate to our peers which messages we are willing to receive and only send messages to peers they are expecting to receive. this is done with "neighbor packets" in the existing gossip code, which are special messages used to establish data-dependency, as @infinity0 frames it.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -989,6 +989,67 @@ Dispatch a `PovFetchSubsystemMessage(relay_parent, candidate_hash, sender)` and
...
+
+On `StatementGossipSubsystemMessageIn::StatementToGossip`:
+1) Get the associated sender for the `relay_parent` hash.
+2) Send the signed statement on it.
+
+```rust
+async fn statement_gossip_job<GS: GossipService>(
+    gossip_service: GS,
+    // The relay parent that this job is running for.
+    relay_parent: Hash,
```
burdges, line 1032: I'd expect:  A backing attestation would've relay_parent matching the candidate receipt I guess.  An availability attestation would've relay patent when the parablock gets declared backed, so well after the relay_parent of the candidate receipt.  An approval/validity attestation would've relay patent when the parablock gets declared available, even further after the relay_parent of the candidate receipt.

docs/polkadot/implementors-guide/guide.md
```diff
@@ -1085,6 +1146,22 @@ enum CandidateBackingSubsystemMessage {
...
+  StatementToGossip { relay_parent: Hash, statement: SignedStatement }
+}
+```
+
+```rust
+enum StatementGossipSubsystemOut {
+  /// A statement that we've received from a validator.
+  StatementReceived { relay_parent: Hash, statement: SignedStatement }
+}
+```
```
burdges, line 1163: Approval VRF gossiping can sign its gossip messages with their own VRFs.  Adds some code strangeness, but saves 64 bytes and 35ish us per gossip messages, maybe not worth the trouble.

docs/polkadot/implementors-guide/guide.md
rphmeier: Yes on the backing statements, although this module is not responsible at the moment for availability or secondary check gossip in any way. Availability bitfields could reasonably be merged with this module, but secondary check results probably will use a separate mechanism as they are likely to just be transactions circulated by the transaction queue gossip.

burdges: I see, so those happen elsewhere probably

docs/polkadot/implementors-guide/guide.md
mxinden: I would imagine the `StatementGossipSubsystem` to own a [`GossipEngine`](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/bridge.rs#L37) as well as some means of communication to a custom implementation of the [`Validator`](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/validator.rs#L23) trait. I don't have an opinion whether the `GossipEngine` should be wrapped in a `GossipService`.

> - communicate to our peers which messages we are willing to receive and only send messages to peers they are expecting to receive. this is done with "neighbor packets" in the existing gossip code, which are special messages used to establish data-dependency, as Ximin frames it.

Sounds good to me. Only thing that would be important to me is that we don't misuse the notion of a `Validator` to send out messages like we do in `substrate/client/finality-grandpa`.

rphmeier: As I understand the gossip model, it's the responsibility of the validator to re-broadcast messages that have been received. Without doing that, messages will not propagate. It's necessary to do that at the gossip-validator level so we can distinguish between neighbor packets and gossip messages. So without changing the lower-level gossip APIs substantially, we have to send messages from the gossip-validator. Or did you mean something different?

mxinden: > As I understand the gossip model, it's the responsibility of the validator to re-broadcast messages that have been received. Without doing that, messages will not propagate. 

Yes, even though I would like to be a bit nit picky and rephrase it as: "A validator validates gossip messages and decides whether a message (if and only if valid) should be further propagated (via the [`ValidationResult`](https://github.com/paritytech/substrate/blob/master/client/network-gossip/src/validator.rs#L76))."

A gossip validator only reacts to messages but is never proactive as in sending new messages (not like what we do today with new neighbor messages in grandpa). Sending new messages (e.g. neighbor messages or statements) should happen through the `GossipEngine`.

rphmeier:
Can you add sections with information on:
  - Message types
  - Under which circumstances we will accept a message
  - Under which circumstances we will send a message to a peer

Also, and this is beyond the scope of the section, but we should provide some context and describe the components that make up a gossip protocol. We're going to have a few different sections that each define gossip protocols, so having some common reference point for what we mean is important for readability. 
https://github.com/w3f/research/blob/master/docs/polkadot/networking/0-overview.md is one link that you could reference (looks better in HackMD). The propagation pool section describes substrate gossip in an abstract way that we can reference or paraphrase here.

PR #95, opened 2020-08-12
XCMP/SPREE Diagram

lamafab:
Preview:

image

FYI: You should probably expect some upcoming changes.


lamafab:
It was mentioned that this shouldn't be here, but to include it in the spec (https://github.com/w3f/polkadot-spec/pull/287).

PR #96, opened 2020-08-24
Update Overview PDF to lastes version 

FlorianFranzen:
Currently the overview paper is not build from source for the website, but it has to be committed to the repo as PDF instead.

This PR updates this PDF to the latest version. 

Generated on NixOS with texlive 2019.


infinity0:
I'd suggest the news page link to the arxiv page instead, and we formalise a release process for updates to the overview paper, with git tags & actual version numbers.

FlorianFranzen:
@infinity0 I would prefer that solution too, and I am setting up something very similar for the spec, including the automated release of PDF diffs between versions. It should be easily portable to this repo once it is done.

PR #107, opened 2020-11-30
networking: a&v: add more detail as requested

infinity0:
Added some more details as requested by @rphmeier - please review


infinity0:
Merging so I can proceed with some more work on top of this, but LMK if you'd like more changes.

PR #111, opened 2020-12-07
added beh. economics section

jonasW3F:
Restructured the "Economics" section to utilize the new left-hand structure. By doing so, I added the "Token Economics" and "Parachain Allocation" chapters to it. In addition, I created a new section for "Behavioral Economics" and added the existing "Validator Selection" write-up to it as well as the new (WIP) section on parachain auction experiment.

Since I am not sure if this is the correct way I did it, I made a PR and request a review.

Thanks ximin!


infinity0:
To update others: I think this looks good, although perhaps we should retain one single top-level section for both behavioural economics and economics - my reasoning is that we'd potentially want to apply behavioural economics methods to all those different topics later, e.g. the topics are associated conceptually with economic methods in general, whether we apply behavioural economics or not to a particular topic is more due to us temporarily deciding where best to spend efforts, which is not very obvious to a reader.

jonasW3F:
> To update others: I think this looks good, although perhaps we should retain one single top-level section for both behavioural economics and economics - my reasoning is that we'd potentially want to apply behavioural economics methods to all those different topics later, e.g. the topics are associated conceptually with economic methods in general, whether we apply behavioural economics or not to a particular topic is more due to us temporarily deciding where best to spend efforts, which is not very obvious to a reader.

I agree. Future economic write-ups will probably incorporate ideas from beh. economics and then having it within one makes sense. Apart from that I'd like to have a place for the specific research / projects I am currently working on. The current commit has a single top-level for both topics and a slightly less nice-looking second level for beh. econ. I think that is fine for know. Let's discuss later in the doc. sync.

PR #114, opened 2021-01-14
Add latest spec and spec team members

FlorianFranzen:
This PR changes the following:

 - Link host-spec directly to [spec release cache](https://w3f.github.io/polkadot-spec)
 - Delete host-spec hosted in research repo
 - Add page for runtime spec (linking to release cache as well)
 - Add spec team to members page
 - Add specs to publication page and remove TODO

Right now @lamafab and my page only contain our names and email addresses. I will it to further PRs to update these.

This a first step towards w3f/polkadot-spec#332 and fixes w3f/polkadot-spec#28. The release cache was implemented in w3f/polkadot-spec#348 and is the basis for this.

This depends on #115 for the embedded PDF to be displayed, as the currently hosted version is still limited to mozillas GitHub Pages.


docs/polkadot/Polkadot-Host-Spec.md
```diff
@@ -15,4 +15,5 @@ presented in a more readable way in preceding documents.
 
 TODO: move to publications
 
-<iframe src="../_static/pdfview/viewer.html?file=../pdf/polkadot-host-spec.pdf" width="100%" height="650em"></iframe>
+<iframe src="../_static/pdfview/viewer.html?file=https://w3f.github.io/polkadot-spec/spec/host/latest.pdf" width="100%" height="650em"></iframe>
```
infinity0, line 18: Did you test that this actually works? I seem to remember some cross-domain restriction or maybe it was just because I was using file:///

docs/team_members.rst
```diff
@@ -0,0 +1,38 @@
...
+
+## Research Team 
+
+The core Web3 Foundation research team working on various areas relevant to decentralised systems.
+
+.. toctree::
+   :glob:
+   :maxdepth: 1
+
+   team_members_research/alfonso
```
infinity0, line 13: Would it be more future-proof to just have everyone be listed under the same directory `team_members/$name`, and only have the structure be expressed on this page? Syed is also in both teams, so that would be less confusing that way.

docs/polkadot.rst
```diff
@@ -16,3 +16,4 @@ Polkadot
    polkadot/economics
    polkadot/security
    polkadot/Polkadot-Host-Spec
+   polkadot/Polkadot-Runtime-Spec
```
infinity0, line 19: The existing links are all top-level sections. So it might be better either to put this under a "specifications" directory, or just have 1 specifications page with multiple PDFs linked on it.

docs/team_members.rst
FlorianFranzen: As I did not add the glob instruction together with the `team_members_research/*` entry, I did not feel like I should change that. But I agree that the mix of globbing and hard-coded is rather ugly. I see two options:

- No hard coding (apart from Syed), two separate folders resulting in an alphabetic list. Might require all file names being lower case.
- Only hard-coded paths, just one folder as suggested.

docs/polkadot.rst
FlorianFranzen: Good point, I will turn this into a specification section instead (i.e. subfolder)

docs/polkadot/Polkadot-Host-Spec.md
FlorianFranzen: Ok, damn. It is not that easy. Will get back to you on this.
```
Message: file origin does not match viewer's
```

FlorianFranzen: This seems to be more a limit of the viewer. I will just switch to a direct embedding, which does not seem to cause that issue for me:
```
<iframe src="https://w3f.github.io/polkadot-spec/spec/host/latest.pdf" width="100%" height="650em"></iframe>
```

docs/team_members.rst
infinity0: Oh right yes I forgot about the glob, but I think it's OK to drop that too. I think sphinx will warn if there are any orphan pages underneath team_members/ that are not referenced by any other pages.

docs/polkadot/Polkadot-Host-Spec.md
infinity0: It might be possible to configure the view to allow some whitelisted other domains, I'll have a look into it..

infinity0: (the error messages comes from the viewer's own js, you can see it in the git repo, so it is probably possible)

infinity0: @FlorianFranzen try this

```diff
diff --git a/docs/_static/pdfview/viewer.js b/docs/_static/pdfview/viewer.js
index 4c65e33..2aaaa9a 100644
--- a/docs/_static/pdfview/viewer.js
+++ b/docs/_static/pdfview/viewer.js
@@ -1546,7 +1546,7 @@ var PDFViewerApplication = {
 };
 var validateFileURL = void 0;
 {
-  var HOSTED_VIEWER_ORIGINS = ['null', 'http://mozilla.github.io', 'https://mozilla.github.io'];
+  var HOSTED_VIEWER_ORIGINS = ['null', 'http://mozilla.github.io', 'https://mozilla.github.io', 'https://w3f.github.io'];
   validateFileURL = function validateFileURL(file) {
     if (file === undefined) {
       return;
```

FlorianFranzen: What is the benefit of hosting our own viewer and not relying on each browsers own PDF viewer?

infinity0: I actually don't know, "it was like that when I got here". We can chat a bit about it in the sync meeting tomorrow?

infinity0: For reference, it was added in #14. Both of them (federico, peter) have left W3F now...

FlorianFranzen: Ok, it seems I am spoiled by Firefox, which just has pdf.js build in. I update the local version instead and replace the mozilla with the w3f github pages subdomain (in accordance with mozilla/pdf.js#6916).

docs/polkadot.rst
infinity0: You forgot to update this to just polkadot/specifications (and rm the previous line with Host-Spec) but otherwise it looks good!

infinity0 (approved):

PR #115, opened 2021-01-19
Update pdf.js and correct domain filter

FlorianFranzen:
This PR update pdf.js to the latest stable release and changes the domain filter to the correct subdomain, as suggested in mozilla/pdf.js#6916.


infinity0 (approved):
LGTM, thanks! I'd also drop the http://w3f.github.io but it's probably not a big deal.

PR #116, opened 2021-01-21
overview paper: deprecate sentry nodes - see also paritytech/substrate#6845

infinity0:
As pointed out by someone on Element


burdges:
I'm fine removing them here, but I'll remark that naive operational security measures, like sentries or remote signing, imposed overhead upon our core goals.  We'll redevelop appropriate operational security once the core protocol stabalizes, so say after 2022.  

Aside from parachains, we'll want all of contextual execution, Sassafras with mempool-free opaeration, XCMP, QUIC with Noise IK etc., multi/hierarchical relay chains, including state archival and lazy code retrieval via availability, before reconsidering sentry nodes.

PR #117, opened 2021-01-21
pdf.js: rewrite validateFileURL to use file whitelist

FlorianFranzen:
As it turns out I misunderstood the purpose of `validateFileURL` in my previous PR, so this changes the function to an actual whitelist of external pdf resources.

- Disables use of `defaultURL` if no file is supplied
- Disables use of `blob:` protocol as file source
- Only viewer origin that disables check now is when run locally (i.e. when developing)
- All non-local pdf files are checked against whitelist

### DIFF
--- a/docs/_static/pdfview/viewer.js
+++ b/docs/_static/pdfview/viewer.js
@@ -2161,7 +2161,8 @@ var PDFViewerApplication = {
 exports.PDFViewerApplication = PDFViewerApplication;
 var validateFileURL;
 {
-  var HOSTED_VIEWER_ORIGINS = ["null", "http://mozilla.github.io", "https://mozilla.github.io"];
+  var HOSTED_VIEWER_ORIGINS = ["null"];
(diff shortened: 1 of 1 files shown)


infinity0:
Normally, this would be a great change - however we are editing a third-party file so we should be keeping the patch diff small and ideally submitting it back to the original project as a PR, which will keep maintenance costs minimal next time you or someone else updates the file (e.g. like we did in #115). Sorry if this seems picky but it is quite standard Best Practises.

To that end, can you please restore the original functionality of HOSTED_VIEWER_ORIGINS and also submit it as a patch to the upstream project? (They will also likely ask you to break down the PR into multiple commits - I see bugfixes in here as well as some minor refactoring, which is positive as a whole but we don't want to carry this by ourselves in W3F.)

If they do not accept your PR, or only accept it partially, we should put the remaining unaccepted stuff into an extra viewer.js.patch file so it is obvious to the next person who touches it (including you in 6 months who might have forgotten about it).

PR #123, opened 2021-10-07
to fix misprint in 2-token-economics.md

agryaznov:
misprint fixed


FlorianFranzen (approved):
Good catch. Thank you for your contribution!

PR #126, opened 2022-01-12
Provide shell.nix to provide build env.

eskimor:
Useful for nix users to quickly get a working build environment.


burdges:
Thanks!

PR #130, opened 2022-05-18
Fix formula for transaction fees

maltekliemann:
This PR removes the unnecessary brackets from the first appearance of the formula for transaction fees and fixes an incorrectly placed bracket in the second appearance of the formula.

Here's a link to the correct formula in the code: https://github.com/paritytech/substrate/blob/b75a253f148aa36fa17cf795b9f2fc2f22d0fcc5/frame/transaction-payment/src/lib.rs#L508-L533


FlorianFranzen (approved):
Nice catch, indeed we should make sure we recall our own definitions correctly. Thank you for your contribution.

PR #145, opened 2023-10-04
Tokenomics update

jonasW3F:
Updated the inflation model and clarified a lot of things.


AlfonsoCev (approved):
It looks good to me. I just removed some references to the staking rates in other blockchain projects, as they felt unnecessary.

